PR #3, opened 2025-11-18
[WF-206,WF-211] Add charm structure + implement airflow coordinator provides

shayancanonical:
## Description
- Adds basic charm code structure (including airflow coordinator provides, airflow config generator)
- Implements a simple charm that has logic for the provides part of the airflow coordinator relation
- Implements scenario tests for added code
- Includes a very minimal airflow config template
- A charm lib that implements classes to be used to encapsulate the coordinator relation for both the requires and provides charms

## Notes
Very minimal PR that introduces basic code structures + pre-flight checks for related airflow core charms
Many portions of introduced code will need to be modified in follow up PRs in the coming weeks

Relevant diagram to understand charm lib code: https://github.com/Gu1nness/charm-ddd/blob/main/2025-05-16_12-19.png
- Databags (from both the requirer and provider sides) are represented as pydantic models
- Custom Airflow Coordinator events take said pydantic models and present them as properties (so they are surfaced easily to the event handler through `event.content`)
- `data_interfaces.EventHandlers` implements all wrapper that provides access to a way to compute diff of databag contents from the previous version -> the old databag content is stored transparently in the databag (on the requires side) in key `data`
- `AirflowCoordinatorRequirerEventHandler` and `AirflowCoordiantorProviderEventHandler` instantiate a `data_interfaces.OpsRelationRepositoryInterface` object (in actuality a `OpsRelationRepository` and `OpsRepository`) which has all the logic for cleanly marshalling and unmarshalling a pydantic model into databag + secrets as appropriate

To summarize: databags + secrets represented as pydantic models, event handlers contain logic for diff computation and custom event emission, repositories contain logic to convert pydantic models <-> databag + secret


src/charm.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def __init__(self, framework: ops.Framework):
+        super().__init__(framework)
+
+        self._config_generator = config_generator.AirflowConfigGenerator(self)
+        self._provider = airflow_coordinator.AirflowCoordinatorProvides(
+            self, AIRFLOW_COORDINATOR_RELATION_NAME
+        )
+
+        self.framework.observe(self.on.start, self.reconcile)
```
jananisenthilkumar291: I'm trying to understand the event flow here. `reconcile` is currently triggered only on `start` and `update_status`. But `start` may fire before leadership is granted, and if the unit isn't leader yet, reconcile exits early. At that moment `_provider` will also see no relations.

Would it make sense to observe `leader_elected` and the relevant relation events (`relation_created / relation_joined/ relation_changed`)? Otherwise the charm might miss the transitions where the unit becomes leader or relations become available, and we’d only pick that up on the next update_status trigger.

templates/airflow_config.j2
```diff
@@ -0,0 +1,13 @@
+[core]
+dags_folder = dags
+executor = LocalExecutor
+load_examples = False
+
+[database]
+sql_alchemy_conn = sqlite:///opt/airflow/airflow.database
+
+[webserver]
+web_server_port = 8080
```
jananisenthilkumar291: As per [Airflow 3.x release](https://airflow.apache.org/docs/apache-airflow/stable/release_notes.html#:~:text=%5Bwebserver%5D%20web_server_port%20%E2%86%92%20%5Bapi%5D%20port), should this rather be?
```
[api]
port = 8080
```

jananisenthilkumar291 (changes requested):

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
```diff
@@ -0,0 +1,595 @@
...
+    "dag-processor",
+]
+
+
+class AirflowCoordinatorRequirerModel(data_interfaces.BaseCommonModel):
+    """Requirer side of the Airflow Coordinator model."""
+
+    airflow_version: str
+    workload_image_hash: str
+    component: typing.Literal["scheduler", "api-server", "triggerer", "dag-processor"]
```
shayancanonical: use `REQUIRED_AIRFLOW_CORE_COMPONENTS` instead of repeating strings (to make it easier to maintain and update)

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
```diff
@@ -0,0 +1,595 @@
...
+
+    validation_failure: dict | None = pydantic.Field(default_factory=dict)
+
+    @pydantic.model_validator(mode="after")
+    def validate_fields(self):
+        """Validates and modifies, if necessary, response to be sent from Airflow Coordinator."""
+        if self.validation_failure:
+            if not all(
+                key in self.validation_failure and self.validation_failure[key]
+                for key in ["failed_component", "failure_code", "error_message"]
```
shayancanonical: suggestion: use a namedtuple to encapsulate a validation failure

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
```diff
@@ -0,0 +1,595 @@
...
+            # TODO: confirm how to remove this secret id or if it can be reused without deleting
+            self.secret_id = None
+
+            return self
+
+        if not self.config_template:
+            raise ValueError("Missing config template")
+
+        # TODO: uncomment once we have a valid pod spec to write in the relation
+        # if not self.kubernetes_executor_pod_spec:
```
shayancanonical: add validation that if `kubernetes_executor_pod_spec` is set, but `config_template` is not -> the `kubernetes_executor_pod_spec` is set to empty with appropriate warning/error logs

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
```diff
@@ -0,0 +1,595 @@
...
+    def _dispatch_events(
+        self,
+        event: ops.RelationEvent,
+        _diff: data_interfaces.Diff,
+        content: AirflowCoordinatorProviderModel,
+    ):
+        if "validation-failure" in _diff.added or "validation-failure" in _diff.changed:
+            getattr(self.on, "airflow_core_metadata_validation_failed").emit(
+                event.relation, app=event.app, unit=event.unit, content=content
+            )
```
shayancanonical, line 348: add return?

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
```diff
@@ -0,0 +1,595 @@
...
+                sensitive_data=sensitive_data,
+            )
+
+        if config_template:
+            model.config_template = config_template
+
+        if kubernetes_executor_pod_spec:
+            model.kubernetes_executor_pod_spec = kubernetes_executor_pod_spec
+
+        if sensitive_data:
```
shayancanonical: todo: how to specify keys in `sensitive_data` that need to be popped from the secret?

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
```diff
@@ -0,0 +1,595 @@
...
+        )
+        self.framework.observe(charm.on[relation_name].relation_broken, callback)
+
+    @property
+    def config(self) -> typing.Optional[str]:
+        """The Airflow config for the core charm to use."""
+        if not self._relation or not self._relation.active:
+            return
+
+        provider_content = self.provider_content
```
shayancanonical: `self.provider_content` -> `self._requirer_handler.provider_content`

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
```diff
@@ -0,0 +1,595 @@
...
+            context=provider_content.sensitive_data
+        )
+
+    @property
+    def kubernetes_executor_pod_spec(self) -> typing.Optional[str]:
+        """The K8s executor pod spec for the core charm to use."""
+        if not self._relation or not self._relation.active:
+            return
+
+        provider_content = self.provider_content
```
shayancanonical: `self.provider_content` -> `self._requirer_handler.provider_content`

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
```diff
@@ -0,0 +1,595 @@
...
+    @property
+    def all_required_components_valid(self) -> tuple[bool, str]:
+        """Check validity of all related core charms."""
+        core_charms_metadata = self._provider_handler.core_charms_metadata
+
+        missing_components = sorted(
+            set(REQUIRED_AIRFLOW_CORE_COMPONENTS) - set(core_charms_metadata.keys())
+        )
+        if missing_components:
+            return False, f"Missing integrations with {', '.join(missing_components)}"
```
shayancanonical: possible to use namedtuple with proper field setting instead of returning a tuple

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
```diff
@@ -0,0 +1,595 @@
...
+        """Update config with related core charms."""
+        self._provider_handler.update_content(
+            config_template=config_template, sensitive_data=sensitive_data
+        )
+
+    def set_kubernetes_executor_pod_spec(
+        self, k8s_executor_pod_spec_template: str, sensitive_data: dict[str, str]
+    ) -> None:
+        """Update k8s executor pod spec with related core charms."""
+        self._provider_handler.update_content(
```
shayancanonical, line 923: validation and error logging required if `config` is empty?

should the two methods be unified (to make sure a config is always provided, even if it has not changed, when the k8s executor pod spec is updated)?

templates/airflow_config.j2
shayancanonical: will update. but for now, this template is treated just as a mock. we will iterate on this (with a dedicated ticket) to make this a valid airflow config

src/charm.py
shayancanonical: good point! just used `start` as a stub, but we need to be careful what events to observe

for `relation_created / relation_joined / relation_changed` -> will update the `AirflowCoordinatorProvides` to observe so all the charm class (`AirflowCoordinatorK8sOperatorCharm) needs to do is instantiate `AirflowCoordinatorProvides`

shayancanonical: now observing `airflow_core_metadata_available` event in the `AirflowCoordinatorProvides` -> implemented in ce231526735e2bdb50bd18be7c8c47beff4c5ed9

templates/airflow_config.j2
shayancanonical: updated in ce231526735e2bdb50bd18be7c8c47beff4c5ed9

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
shayancanonical: updated in ce231526735e2bdb50bd18be7c8c47beff4c5ed9

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
shayancanonical: using pydantic base model instead of namedtuple -> ce231526735e2bdb50bd18be7c8c47beff4c5ed9

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
shayancanonical: not possible to set k8s pod spec without config (as config is required)

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
shayancanonical: added in ce231526735e2bdb50bd18be7c8c47beff4c5ed9

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
shayancanonical: changed design so that sensitive_data will not be an `.update()` but rather the sensitive data provided will overwrite the previous version of the sensitive data entirely

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
shayancanonical: addressed in ce231526735e2bdb50bd18be7c8c47beff4c5ed9

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
shayancanonical: addressed in ce231526735e2bdb50bd18be7c8c47beff4c5ed9

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
shayancanonical: returning only an optional string -> ce231526735e2bdb50bd18be7c8c47beff4c5ed9. will consider if a better capsule available to surface to the charm (e.g. `ExceptionWithStatus`)

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
shayancanonical: merged the two methods -> ce231526735e2bdb50bd18be7c8c47beff4c5ed9

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
```diff
@@ -0,0 +1,836 @@
+# Copyright 2025 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+# TODO: move code to officially generated charm lib once airflow-coordinator-k8s
+# is registered and first revision published on charmhub
```
DnPlas: I don't think we need to wait for this to happen, as long as the charm is registered, the library can also be registered. I wouldn't wait too much on this.

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
```diff
@@ -0,0 +1,836 @@
...
+        if not relation:
+            raise ValueError("Missing relation")
+
+        self.relation = relation
+
+        app_name = snapshot.get("app_name")
+        self.app = self.framework.model.get_app(app_name) if app_name else None
+
+        unit_name = snapshot.get("unit_name")
+        self.unit = self.framework.model.get_app(unit_name) if unit_name else None
```
DnPlas: typo?

```suggestion
        self.unit = self.framework.model.get_unit(unit_name) if unit_name else None
```

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
```diff
@@ -0,0 +1,836 @@
...
+
+class AirflowCoordinatorRequires(ops.Object):
+    """A requirer handler encapsulating the airflow coordinator relation."""
+
+    def __init__(
+        self,
+        charm: ops.CharmBase,
+        relation_name: str,
+        component: str,
+        workload_container_name: str,
```
DnPlas: Would it be a better idea to just pass the Container object, most if not all charms will already be getting this in their code?

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
```diff
@@ -0,0 +1,836 @@
...
+        self._charm = charm
+        self._component = component
+
+        self._relation = charm.model.get_relation(relation_name)
+
+        self.workload_container = charm.unit.get_container(workload_container_name)
+        if self._relation and self.workload_container.can_connect():
+            # TODO: pull airflow_version and workload_image_hash from container
+            airflow_version = "3.1.0"
+            workload_image_hash = "somehash"
```
DnPlas, line 674: Please link https://github.com/canonical/airflow-rocks/issues/13 here so we don't forget

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
```diff
@@ -0,0 +1,836 @@
...
+                _build_requirer_model(relation)
+                for relation in self.interface.relations
+                if self.interface.repository(relation.id, relation.app).get_data()
+            ]
+            if metadata is not None
+        }
+
+
+class AirflowCoordinatorRequires(ops.Object):
+    """A requirer handler encapsulating the airflow coordinator relation."""
```
DnPlas, line 645: Let's improve docstrings to help consumers of this lib how to use the requires.

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
```diff
@@ -0,0 +1,836 @@
...
+        callback: typing.Callable,
+    ):
+        if component not in REQUIRED_AIRFLOW_CORE_COMPONENTS:
+            raise ValueError(f"Invalid component {component}")
```
DnPlas: If we already have the `AirflowCoordinatorRequirerModel`, would it make sense to let pydantic do this check?


```suggestion
        AirflowCoordinatorRequirerModel(
            airflow_version="3.1"
            workload_image_hash="somehash",
            component=component, # <-- check the required component
        )
```

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
```diff
@@ -0,0 +1,836 @@
...
+        self._requirer_handler = AirflowCoordinatorRequirerEventHandler(
+            charm, relation_name, AirflowCoordinatorProviderModel
+        )
+
+        self._charm = charm
+        self._component = component
+
+        self._relation = charm.model.get_relation(relation_name)
+
+        self.workload_container = charm.unit.get_container(workload_container_name)
```
DnPlas: Do these two make more sense as properties?

DnPlas: Also, `get_container` will raise "if the named container doesn’t exist", potentially you want to add some handling or let the charm do it and just pass the Container obj (same msg from above).

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
```diff
@@ -0,0 +1,836 @@
...
+        if self._relation and self.workload_container.can_connect():
```
DnPlas: What is this check doing? Is this temporal?

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
```diff
@@ -0,0 +1,836 @@
...
+            failure.message
+            for failure in self._requirer_handler.validation_failures
+            if failure.component == self._component
+        ]
+
+    @property
+    def missing_core_components_exist(self) -> bool:
+        """Indicates if coordinator reports missing components for the cluster."""
+        if not self.relation_present:
+            return True
```
DnPlas: These two lines are not doing exactly what the docstring says, but rather just returning True on missing relation.

If we want to report that the coordinator is missing components, should we just remove this check? It should be covered by the return from below, isn't?

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
```diff
@@ -0,0 +1,836 @@
...
+            if failure.code == MISSING_COMPONENT
+        )
+
+    def write_airflow_config(self, config_path: str) -> bool:
+        """Render the Airflow config in the provided path in the workload container."""
+        if not self.workload_container.can_connect():
+            return False
+
+        if not self.ready:
+            return False
```
DnPlas: I think these checks should be done on the requirer side, they seem off here. When I was consuming this library, it felt unnatural that I had to check for a `True` from this method, while it is actually executing something.

I think it'd be better to raise here and let the requirer side handle that, rather. In the charm code we'd need to handle a `PathError`. wdyt?

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
```diff
@@ -0,0 +1,836 @@
...
+            config_path,
+            config,
+            user="root",
+            group="root",
+            make_dirs=True,
+        )
+
+        return True
+
+    def write_kubernetes_executor_pod_spec(self, filepath: str) -> bool:
```
DnPlas: Similar case to the previous method.

src/charm.py
```diff
@@ -0,0 +1,55 @@
...
+    def reconcile(self, _) -> None:
+        """Idempotent reconcile method to handle most relevant charm events."""
+        # TODO: restrict the application to 1 unit?
+        if not self.unit.is_leader():
+            return
+
+        error_message = self._provider.validate_core_components()
+        if error_message:
+            self.unit.status = ops.BlockedStatus()
+            self.app.status = ops.BlockedStatus(error_message)
```
DnPlas: If this charm is supposed to have one unit only, why are we setting the status both at unit and app level?

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
```diff
@@ -0,0 +1,836 @@
...
+
+logger = logging.getLogger(__name__)
+
+REQUIRED_AIRFLOW_CORE_COMPONENTS_LITERAL = typing.Literal[
+    "scheduler",
+    "api-server",
+    "triggerer",
+    "dag-processor",
+]
+REQUIRED_AIRFLOW_CORE_COMPONENTS = list(typing.get_args(REQUIRED_AIRFLOW_CORE_COMPONENTS_LITERAL))
```
DnPlas: I'm a bit confused as to why we have to do this, perhaps is to make it easier to iterate.

I wonder if we can make all this less convoluted and instead use pydantic fully. In the sense of, allow it to make the checks for us in the various places where we do something like:

```
        if self.component not in REQUIRED_AIRFLOW_CORE_COMPONENTS:
            raise ValueError(f"Invalid component {self.component}")
```

This would potentially mean a small refactor in your code, but I think it'd be less convoluted. I'll leave it up to you.

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
```diff
@@ -0,0 +1,836 @@
...
+    def validate_fields(self):
+        """Validates that no inconsistent request sent to Airflow Coordinator."""
+        if not self.airflow_version:
+            raise ValueError("Missing airflow version")
+
+        if not self.workload_image_hash:
+            raise ValueError("Missing workload image hash")
+
+        if not self.component:
+            raise ValueError("Missing component")
```
DnPlas: I believe this check is already covered by Pydantic.

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
```diff
@@ -0,0 +1,836 @@
...
+        """Validates and modifies, if necessary, response to be sent from Airflow Coordinator."""
+        if self.validation_failures:
+            return self
+
+        if not self.config_template:
+            raise ValueError("Missing config template")
+
+        # TODO: uncomment once we have a valid pod spec to write in the relation
+        # if not self.kubernetes_executor_pod_spec:
+        #     raise ValueError("Missing kubernetes executor pod spec")
```
DnPlas: Let's create a task or GH issue to track this and add a comment. Let's not forget about it.

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
```diff
@@ -0,0 +1,836 @@
...
+        """Builds an Airflow Coordinator requirer event handler."""
+        super().__init__(charm, relation_name, unique_key)
+        self.charm = charm
+        self.component = self.charm.app
+        self.request_model = request_model
+        self.interface = data_interfaces.OpsRelationRepositoryInterface(
+            charm.model, relation_name, request_model
+        )
+
+        self.relation = self.charm.model.get_relation(relation_name)
```
DnPlas, line 330: I think this is the same as in other places, we may want to just pass the Relation object to the Provider class, that way we prevent issues later if this is `None` and ditch a lot of checks in this lib.

charmcraft.yaml
```diff
@@ -0,0 +1,44 @@
+# This file configures Charmcraft.
+# See https://documentation.ubuntu.com/charmcraft/stable/reference/files/charmcraft-yaml-file/
+type: charm
+name: airflow-coordinator-k8s-operator
```
DnPlas:
```suggestion
name: airflow-coordinator-k8s
```

charmcraft.yaml
```diff
@@ -0,0 +1,44 @@
...
+    plugin: uv
+    source: .
+    build-snaps:
+      - astral-uv
+  files:
+    plugin: dump
+    source: .
+    stage:
+      - LICENSE
+      - templates
```
DnPlas: Should we place templates inside `src/` and avoid this dump? I know most kf charms do it this way. Not a biggie, tho.

DnPlas (commented):
Thanks @shayancanonical !

I just went through some of the general stuff, but couldn't get to the tests.

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
shayancanonical: moved to a legal charm lib file in 29a021a223926efea85ef287c82ae0d29f6bd3a8

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
shayancanonical: addressed in 9fee0f46b1e599a7112889db01941e444cdc84cf

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
shayancanonical: great suggestion. now passing in container starting 9fee0f46b1e599a7112889db01941e444cdc84cf

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
shayancanonical: added link to issue in 9fee0f46b1e599a7112889db01941e444cdc84cf

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
shayancanonical: simplified all validators to eliminate redundant validations + using enums where it makes sense (9fee0f46b1e599a7112889db01941e444cdc84cf)

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
shayancanonical: no longer using `get_container()` -> instead passing in the container as a param

additionally, checking if relation exists early in the constructor and noop-ing if it does not exist -> this simplifies the need to check for false-y relations

9fee0f46b1e599a7112889db01941e444cdc84cf

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
shayancanonical: the check existed to make sure we only set metadata in the relation once:
1. relation exists -> due to previous comment, the constructor early-outs if relation does not exist. this check no longer needed
2. workload container's pebble needs to be reachable to be able to query the airflow version + retrieve the image hash

2) is a little early, but will be needed as soon as we add a hash in the rock

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
shayancanonical: refactored based on offline discussions in 9fee0f46b1e599a7112889db01941e444cdc84cf

essence of refactor: separating properties/methods that are purely checks (and return bools) from those that actually perform some action/changes (like writing a file)

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
shayancanonical: refactored based on offline discussions starting in 9fee0f46b1e599a7112889db01941e444cdc84cf

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
shayancanonical: addressed in 9fee0f46b1e599a7112889db01941e444cdc84cf

src/charm.py
shayancanonical: it was a spur of the moment code addition while testing relations in a local juju model

now only setting unit status starting 9fee0f46b1e599a7112889db01941e444cdc84cf. we can revisit how to properly ensure unit and app status are accurate as a follow up

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
shayancanonical: addressed in 9fee0f46b1e599a7112889db01941e444cdc84cf. use pydantic's native validations + remove all redundant valdiators

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
shayancanonical: check removed in 9fee0f46b1e599a7112889db01941e444cdc84cf

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator_temp.py
shayancanonical: removed validator to use pydantic's native validation (9fee0f46b1e599a7112889db01941e444cdc84cf). no longer need the link

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
shayancanonical: im not sure if this particular check can be eliminated as it is specific to how events are snapshot-ed and restored by ops. tried my best to eliminate similar relation related checks where possible

charmcraft.yaml
shayancanonical: changed in 9fee0f46b1e599a7112889db01941e444cdc84cf

charmcraft.yaml
shayancanonical: im ok with both approaches. moved in `src/` (9fee0f46b1e599a7112889db01941e444cdc84cf) to stay consistent with charm patterns within analytics

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
shayancanonical: rewrote docstrings in 29a021a223926efea85ef287c82ae0d29f6bd3a8

src/charm.py
```diff
@@ -0,0 +1,71 @@
...
+            self, AIRFLOW_COORDINATOR_RELATION_NAME, callback=self.reconcile
+        )
+
+        for event in [
+            self.on.start,
+            self.on.update_status,
+        ]:
+            self.framework.observe(event, self.reconcile)
+
+    def reconcile(self, _) -> None:
```
DnPlas, line 35: not a blocker - we probably want to follow the same pattern as the core charms, using the custom exception. Until we resolve the part of how to share it, we could leave this as is, but consider it for the future.

DnPlas (approved):
Thanks @shayancanonical ! LGTM

src/charm.py
shayancanonical: i agree - i was planning on following the same pattern for `ExitWithStatusError` in a PR more oriented towards charm logic addition (may be even the PR adding postgres support)

jananisenthilkumar291 (approved):
LGTM! Approved!

PR #4, opened 2025-12-05
[WF-216] Implement integration with postgres as a backend metadata database

shayancanonical:
## Description

- Add integration with postgres database (uses data interfaces v0 lib -> compatible with both postgres 14 and 16 charms + pgbouncer charm)
- Add unit tests

---
## Testing instructions

tox -e unit

Integration tests will be added as a quick followup effort

---
## Checklist (all that apply)
- [ ] Documentation updated


src/charm.py
```diff
@@ -6,49 +6,99 @@
...
-        self.framework.observe(self.on.start, self.reconcile)
-        self.framework.observe(self.on.update_status, self.reconcile)
+        self._database_requires = data_interfaces_v0.DatabaseRequires(
+            self, constants.POSTGRES_RELATION_NAME, database_name=constants.AIRFLOW_DATABASE_NAME
+        )
+        self.framework.observe(self._database_requires.on.database_created, self._reconcile)
+        self.framework.observe(self._database_requires.on.endpoints_changed, self._reconcile)
+        self.framework.observe(
+            self.on[constants.POSTGRES_RELATION_NAME].relation_broken, self._reconcile
+        )
```
DnPlas: tiny comment, tou could do something like:

```
        change_events = [
            self._database_requires.on.database_created,
           self._database_requires.on.endpoints_changed,
            self.on[constants.POSTGRES_RELATION_NAME].relation_broken,
        ]
        for event in change_events:
            self.framework.observe(event, self._reconcile)
```

src/charm.py
```diff
@@ -6,49 +6,99 @@
...
+
+            return False
+
+        if not self._database_requires.is_resource_created():
+            if raise_exceptions:
+                raise ExceptionWithStatusError(
+                    "Waiting for airflow database to be created", ops.WaitingStatus
+                )
+
+            return False
```
DnPlas: Should we rather rely on the exception to be caught and handled in the reconcile method? That way we don't need to add the bool checks, making the code simpler.

src/config_generator.py
```diff
@@ -17,16 +20,45 @@ def __init__(self, charm: ops.CharmBase):
...
+            postgres_relation_id
+        ]
+
+        endpoints = [
+            endpoint for endpoint in relation_data.get("endpoints", "").split(",") if endpoint
+        ]
+        if not endpoints:
+            return None
+
+        return f"postgresql+psycopg2://{relation_data.get('username')}:{relation_data.get('password')}@{endpoints[0]}/{constants.AIRFLOW_DATABASE_NAME}"
```
DnPlas, line 47: Thinking a bit ahead, should we leave this a tad more generic in case we want to integrate with mysql in the future?

src/config_generator.py
```diff
@@ -17,16 +20,45 @@ def __init__(self, charm: ops.CharmBase):
...
+
     @property
     def sensitive_config_values(self) -> dict[str, str]:
         """All sensitive values that will be included in the Airflow config template."""
+        if not self._charm._is_ready(raise_exceptions=False):
+            return {}
```
DnPlas: You'll most likely never see `_is_ready()` returning false here, isn't? Your reconcile method will make sure that if it is not ready, this `sensitive_config_values()` method will never run. Could you confirm?

src/charm.py
```diff
@@ -6,49 +6,99 @@
...
+                return
+        except ExceptionWithStatusError as e:
+            self.unit.status = e.status
             return
 
-        self._provider.set_airflow_config(
+        self._config_provider.set_airflow_config(
             self._config_generator.config_template,
             sensitive_data=self._config_generator.sensitive_config_values,
         )
```
DnPlas: Maybe abstract this to a method that raises ExceptionWithStatusError on potential issues and call it inside the try/except block from before. I see potential issues with uncaught exceptions.

DnPlas (commented):
Thanks @shayancanonical, couple comments.

src/config_generator.py
shayancanonical: currently our charmcraft only supports integration with postgres explicitly

if we add a mysql relation endpoint in the future, i believe we will be able to add conditions to generate a different `_sql_alchemy_connection_string` based on the type of the related database

currently, the sqlalchemy connection string (as enforced by the ORM directly) explicitly requires the database (+ optional connector) to be considered valid

src/charm.py
shayancanonical: great suggestion. implemented in 30f82e28a207834c93094b73f003a5060ed15e84

src/charm.py
shayancanonical: completely relying on exceptions being raised (30f82e28a207834c93094b73f003a5060ed15e84)

assuming that proper checks performed before attempts to generate configs in charm code

src/config_generator.py
shayancanonical: removed in 30f82e28a207834c93094b73f003a5060ed15e84 with a simplified approach (using `ExceptionWithStatusError`)

src/charm.py
shayancanonical: addressed in 30f82e28a207834c93094b73f003a5060ed15e84

src/config_generator.py
DnPlas: Noted

lib/charms/data_platform_libs/v0/data_interfaces.py
DnPlas, line 1: Is this an error? I believe this file has been already added to the repo.

DnPlas (approved):
Thanks @shayancanonical ! LGTM, a tiny note about the lib, though

lib/charms/data_platform_libs/v0/data_interfaces.py
shayancanonical: Actually, the postgres charm (both 14 and 16 tracks) only supports data_interfaces v0

data_interfaces v1: implements some changes to underlying storage (how secrets are used to store data). it is the one that uses pydantic to represent relation data -> we're extending this lib for coordinator + already exists in the repo

data_interfaces v0: the only lib that is supported by postgres charm so far. i recall there being some backwards compatibility (v1 requires can be used with v0 provides -> but i'm not sure and havent tested yet) - i can create a ticket to explore this if we'd like

shayancanonical: update: researched possibility of using requirer with v1 but provider with v0. this is not supported

Issue #6, opened 2025-12-19
Secret is not re-granted after relation broken/departed and re-joined


DnPlas:
### Bug Description

A requirer charm whose relation has been broken by `juju remove-relation coordinator <core charm>` and re-integrated immediately after will never become active because the secret is not re-granted afterwards. This causes the `self.<requires object>.can_write_airflow_config` to never be `True`.

### To Reproduce

1. Deploy all four core charms (from [this branch](https://github.com/canonical/airflow-core-operators/tree/add-airflow-api-server) for instance)
2. Deploy the coordinator (make sure you have the changes from [this other branch](https://github.com/canonical/airflow-coordinator-k8s-operator/tree/feature/add_integration_tests))
3. Integrate all core charms with coordinator
4. Deploy postgresql-k8s
5. Integrate coordinator with postgresql-k8s
6. Remove the relation from a core charm e.g. `juju remove-relation scheduler coordinator`
7. Add the relation back again `juju integrate scheduler coordinator`
8. Observe ALL charms going into waiting status because `can_write_airflow_config` is False

Additionally, check that the secret is granted for all the other core charms, except for the ones whose relation has been removed and re-added.

### Environment

* juju 3.x (LTS)
* k8s v1.32.10

### Relevant log output

```shell
# juju status
scheduler/0*      waiting   idle   10.1.0.219         Waiting for the container or the interface to be ready.
triggerer/0*      waiting   idle   10.1.0.96          Waiting for the container or the interface to be ready.

# secret not granted to scheduler, even if the relation exists

juju show-secret  --reveal d520g3nmp25c762t95m0
d520g3nmp25c762t95m0:
  revision: 1
  checksum: ce487ee6dd3456d0acc9f92fadd0bb281d5bf3a3d23c95a4c96c8a0b5df12b9b
  owner: coord
  label: airflow-coordinator.61.sensitive-data.secret
  created: 2025-12-18T13:56:31Z
  updated: 2025-12-18T13:56:31Z
  content:
    sensitive-data: '{"sql_alchemy_connection_string": "postgresql+psycopg2://relation_id_58:XwE08AsuZlTeqDda@pg-primary.airflow-demo.svc.cluster.local:5432/airflow"}'

# secret granted to api-server

$ juju show-secret  --reveal d520g3nmp25c762t95l0
d520g3nmp25c762t95l0:
  revision: 1
[...]

PR #7, opened 2026-01-07
[WF-212] Implement minimal integration tests

shayancanonical:
## Description

This PR introduces minimal integration tests to assert:
- all charms can be deployed and related
- all relations can be broken and re-formed with expected behavior along the way
- a subset of relations can be broken and re-formed with expected behavior along the way

A new local charm was added for the purposed of mocking airflow core charms

Along the way, https://github.com/canonical/airflow-coordinator-k8s-operator/issues/6 was addressed. The fix was introduced in https://github.com/canonical/airflow-coordinator-k8s-operator/commit/4962e433f834729dca197bd2912aca567f13d235. Changes will be reviewed in #8, but they are included in this PR also to ensure successful execution of the implemented integration tests

CI now includes integration tests to be run upon PR

---
## Testing instructions

Always run tests with `just` as there is always additional set up and clean up included.

`just unit`
`just integration`

---
## Checklist (all that apply)
- [ ] Documentation updated


src/config_generator.py
```diff
@@ -29,22 +25,29 @@ def config_template(self) -> str:
...
+            postgres_relation_id, "password"
+        )
+        database = self._charm._database_requires.fetch_relation_field(
+            postgres_relation_id, "database"
+        )
 
         endpoints = [
-            endpoint for endpoint in relation_data.get("endpoints", "").split(",") if endpoint
+            endpoint
+            for endpoint in self._charm._database_requires.fetch_relation_field(
```
jananisenthilkumar291: Instead of performing four individual lookups with `fetch_relation_field`, should we use a single call to `fetch_relation_data` to retrieve the entire credentials dictionary at once?

If not can we be sure that postgres won't update endpoints or rotate credentials between 2 consecutive calls to `fetch_relation_field`?

shayancanonical: use `fetch_relation_data` instead in f7a2140e4294e7960573a522276f465fe0a83236

to clarify a bit, `fetch_relation_field` access relation databag. a relation databag (in ops) is a snapshot of the databag contents as it existed when the event handler was evoked. thus subsequent calls to `fetch_relation_field` query the snapshot and is assured to be consistent even if the databag changes. the following `relation-changed` event will reflect the updated databag, and our `_reconcile` would handle that subsequently

still valid suggestion, we should just use `fetch_relation_data` to make our code a bit more efficient and clean

jananisenthilkumar291: Understood!

.github/workflows/pull_request.yaml
DnPlas, line 1: So clean!

tests/integration/mock-core-charm/LICENSE
```diff
@@ -0,0 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2025 Shayan
```
DnPlas: You may want to change this to Canonical.
Also, do we need a LICENSE for the mock charm?

tests/integration/test_charm.py
```diff
@@ -12,10 +13,243 @@
...
+    )
+
+    logger.info("Integrating coordinator <-> postgres")
+
+    juju.integrate("airflow-coordinator-k8s", "postgresql-k8s")
+
+    juju.wait(
+        lambda status: jubilant.all_blocked(status, "airflow-coordinator-k8s")
+        and status.apps["airflow-coordinator-k8s"].app_status.message
+        == "Missing integrations with: api-server, dag-processor, scheduler, triggerer"
```
DnPlas: (non-blocking) If we change the message in the charm, this will break. To avoid this easy to miss issue, we can reference the msg if it comes from a global var from the charm code.

This comment applies in different places.

tests/integration/test_charm.py
```diff
@@ -12,10 +13,243 @@
...
+AIRFLOW_COMPONENTS = [
+    "scheduler",
+    "api-server",
+    "triggerer",
+    "dag-processor",
+]
 
-def test_deploy(charm: pathlib.Path, juju: jubilant.Juju):
+
+def test_deploy(juju: jubilant.Juju, charm: pathlib.Path, mock_core_charm: pathlib.Path):
```
DnPlas, line 32: I believe this test case is doing much more than it says. Perhaps you could separate all the assertions on this test case into smaller test cases? E.g. test_deploy only verifies charms go to active after adding required relations, test_db_data can test the actual pgsql data, etc.

DnPlas (commented):
Thanks @shayancanonical, just a few comments.

tests/integration/test_charm.py
shayancanonical: broke the deploy test -> deploy test + relate test in https://github.com/canonical/airflow-coordinator-k8s-operator/pull/7/commits/adbb0586ad789efbd79193c498cfab884c1ea853

will add particular postgres related assertions in the PR that introduces proper handling of postgres relation break

tests/integration/test_charm.py
shayancanonical: good suggestion. updated in https://github.com/canonical/airflow-coordinator-k8s-operator/pull/7/commits/adbb0586ad789efbd79193c498cfab884c1ea853

updated both unit tests and integration tests to use constants when making assertions

tests/integration/mock-core-charm/LICENSE
shayancanonical: oh man, i hate that `charmcraft pack` adds my name in license

removed license from mock charm in https://github.com/canonical/airflow-coordinator-k8s-operator/pull/7/commits/adbb0586ad789efbd79193c498cfab884c1ea853 since it will never be published and there already exists a license in the repo home

.github/workflows/pull_request.yaml
shayancanonical: yay! also plan to implement @jananisenthilkumar291's spec for using concierge + spread to run tests in isolated envs as a separate PR

DnPlas (approved):
Thanks @shayancanonical, LGTM.

jananisenthilkumar291 (approved):
LGTM! Thanks @shayancanonical.

PR #9, opened 2026-01-09
[WF-255] Address postgres relation breaks after everything goes active

shayancanonical:
## Description

This PR implements the recognition + proper handling of postgres relation break after everything is active (configs shared with all related airflow core charms).

Since relation breaks with postgres results in user credential invalidation, all airflow core workloads will be unable to connect to the database. Thus, the lib needs to recognize this happening and signal to the related core charms so that they may halt their workloads.

---
## Testing instructions

`just unit`
`just integration`

This PR sets the base branch as the branch that introduces minimal integration tests. When #7 is merged, this PR will auto set the base branch as `track/3.1`

---
## Checklist (all that apply)
- [ ] Documentation updated


src/charm.py
```diff
@@ -67,9 +70,20 @@ def _all_database_connection_details_present(self) -> None:
             for field in ["username", "password", "endpoints", "database"]
         )
 
+    def _required_dependencies_exist(self) -> bool:
+        """Indicates if all required dependencies for the coordinator exist."""
```
DnPlas: (nit) According to [PEP8](https://peps.python.org/pep-0257/#one-line-docstrings):

>The docstring is a phrase ending in a period. It prescribes the function or method’s effect as a command (“Do this”, “Return that”), not as a description; e.g. don’t write “Returns the pathname …”.

DnPlas (dismissed):
Overall looks good, left a tiny nit. I tested this PR by deploying the coordinator and removing/adding back the relation with pgsql, it all works as expected.

I think we want to set some rules and standards for the core charms, as in my testing, when I removed the relation, all of them went to error instead of waiting. Unrelated to this PR, let's just keep that in mind.

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
jananisenthilkumar291, line 718: nit. [non-blocking]
The comment is probably not relevant to this PR. However, I think the control will not reach here on `relation_broken` due to check in line 690.

jananisenthilkumar291 (dismissed):
LGTM! Left a tiny nit, totally harmless and non-blocking.

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
shayancanonical: added a relation-broken observe in the if statement on line 686 (afc344f00c41deaa25d0f9742365779f785df866)

since while handling the relation-broken event, the relation is not present, the change adds a `self.framework.observe()` for the relation-broken if the relation does not exist. with this, we will call `_reconcile` for coordinator relation broken regardless of whether the relation exists in ops or not

src/charm.py
shayancanonical: updated comment to follow pep8 more closely in afc344f00c41deaa25d0f9742365779f785df866

jananisenthilkumar291 (dismissed):
LGTM! Changes work as expected.

DnPlas (dismissed):
Thanks @shayancanonical !

DnPlas (approved):
I did not see major changes, approving again. Please merge and open a new PR if new changes are required.

PR #12, opened 2026-01-26
feat: run airflow db migrate command

### LINKED ISSUE #11: `airflow db migrate` command missing
### Problem statement

According to the Airflow [docs](https://airflow.apache.org/docs/apache-airflow/stable/installation/setting-up-the-database.html):

> Usually, you need to run airflow db migrate in order to create the database schema if it does not exist or migrate to the latest version if it does.

This translates into one of the charms of the Airflow solution to be responsible to run this on startup and upgrades, and the coordinator seems like the best place to do it.

### Enhancement Proposal

The coordinator charm should execute the `airflow db migrate` command on:

* start, right after integrating with the DB
* Airflow upgrades, right after the upgrade, but before starting the Airflow services because of:

> You should make sure that Airflow components are not running while the database migration is being executed.

Please NOTE that for this to work, we must ensure:

1. The coordinator charm will now have a workload container
2. There must exist an airflow configuration file inside the container

### What needs to get done?

Implement the command executor

DnPlas:
## Description

To ensure all Airflow services can be started, the airflow db migrate
command must run at least once. According to the docs:

Usually, you need to run airflow db migrate in order to create the database schema
if it does not exist or migrate to the latest version if it does.

This commit enables the coordinator charm to do it as it has access to
the Airflow config and the database credentials. It can also control
when the command is run and under which conditions.

For this feature to work, the coordinator charm now has a workload
container that uses the ubuntu/airflow image.

Fixes https://github.com/canonical/airflow-coordinator-k8s-operator/issues/11

---
## Testing instructions

1. Clone the core charms repo and update the library with the version from this PR
2. Build the core and the coordinator charms
3. Deploy postgresql-k8s 14/stable and integrate with coordinator
4. Integrate coordinator with core charms
5. Check debug logs, we are looking for:

```
unit-coor-0: 14:18:22 INFO unit.coor/0.juju-log airflow-coordinator:134: 'airflow db migrate' completed successfully
```

6. The services must be up and running, for example

```
$ kubectl logs -nairflow-demo api-server-0 -c airflow-api-server
2026-01-26T20:18:25.743Z [pebble] GET /v1/notices?after=2026-01-26T20%3A18%3A25.586989401Z&timeout=30s 149.610424ms 200
2026-01-26T20:18:26.051Z [airflow]   ____________       _____________
2026-01-26T20:18:26.051Z [airflow]  ____    |__( )_________  __/__  /________      __
2026-01-26T20:18:26.051Z [airflow] ____  /| |_  /__  ___/_  /_ __  /_  __ \_ | /| / /
2026-01-26T20:18:26.051Z [airflow] ___  ___ |  / _  /   _  __/ _  / / /_/ /_ |/ |/ /
2026-01-26T20:18:26.051Z [airflow]  _/_/  |_/_/  /_/    /_/    /_/  \____/____/|__/
2026-01-26T20:18:26.051Z [airflow] 2026-01-26T20:18:26.051417Z [warning  ] `api_auth/jwt_secret` was empty, using a generated one for now. Please set this in your config [airflow.api_fastapi.auth.tokens] loc=tokens.py:529
2026-01-26T20:18:26.051Z [airflow] 2026-01-26T20:18:26.051675Z [info     ] Running the uvicorn with:
2026-01-26T20:18:26.051Z [airflow] Apps: all
2026-01-26T20:18:26.051Z [airflow] Workers: 1
2026-01-26T20:18:26.051Z [airflow] Host: 0.0.0.0:8080
2026-01-26T20:18:26.051Z [airflow] Timeout: 120
2026-01-26T20:18:26.051Z [airflow] Logfiles: -
2026-01-26T20:18:26.051Z [airflow] ================================================================= [airflow.cli.commands.api_server_command] loc=api_server_command.py:54
2026-01-26T20:18:26.080Z [airflow] /lib/python3.12/site-packages/airflow/api_fastapi/execution_api/routes/__init__.py:23 DeprecationWarning: 'HTTP_422_UNPROCESSABLE_ENTITY' is deprecated. Use 'HTTP_422_UNPROCESSABLE_CONTENT' instead.
2026-01-26T20:18:26.615Z [airflow] /lib/python3.12/site-packages/airflow/api_fastapi/core_api/routes/public/__init__.py:37 DeprecationWarning: 'HTTP_422_UNPROCESSABLE_ENTITY' is deprecated. Use 'HTTP_422_UNPROCESSABLE_CONTENT' instead.
```

---
## Checklist (all that apply)
- [ ] Documentation updated

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -35,13 +35,8 @@ class ExceptionWithStatusError(Exception):
 class AirflowCoordinatorK8SOperatorCharm(ops.CharmBase):
     """Charm the application."""
 
-    _stored = ops.StoredState()
-
     def __init__(self, framework: ops.Framework):
         super().__init__(framework)
-
-        # StoredState to save state of db migration command
-        self._stored.set_default(db_migration_ran=False)
 
         self._container = self.unit.get_container(constants.WORKLOAD_CONTAINER_NAME)
         self._config_generator = config_generator.AirflowConfigGenerator(self)
@@ -112,6 +107,31 @@ class AirflowCoordinatorK8SOperatorCharm(ops.CharmBase):
             },
         }
         return layer
+
+    @property
+    def _peer_relation(self) -> ops.Relation:
+        """Return the peer relation.
+
+        Raises:
+            ExceptionWithStatusError: If peer relation is not available.
+        """
+        peer_relation = self.model.get_relation(constants.PEER_RELATION_NAME)
+        if not peer_relation:
+            raise ExceptionWithStatusError(
+                constants.WAITING_FOR_PEER_RELATION_MESSAGE, ops.WaitingStatus
+            )
+        return peer_relation
+
+    @property
+    def _db_migration_ran(self) -> bool:
+        """Check if database migration has been run."""
+        return self._peer_relation.data[self.app].get("db_migration_ran") == "true"
+
+    @_db_migration_ran.setter
+    def _db_migration_ran(self, value: bool) -> None:
+        """Set database migration state."""
+        # We need to cast the bool to str, Juju relation data bags can only store strings
+        self._peer_relation.data[self.app]["db_migration_ran"] = str(value).lower()
 
     def _required_dependencies_exist(self) -> bool:
         """Returns whether all required dependencies for the coordinator exist."""
@@ -209,10 +229,12 @@ class AirflowCoordinatorK8SOperatorCharm(ops.CharmBase):
             self._configure_pebble_layer()
             self._write_airflow_config()
 
-            # Use StorageState to track if db migration has run before
-            if self._stored.db_migration_ran is False:
+            # Use peer relation data to track if db migration has run
+            # TODO: once we have upgrade logic, we'll need to change the
+            # conditions under which this state will be True/False.
+            if not self._db_migration_ran:
                 self._run_db_migrate()
-                self._stored.db_migration_ran = True
+                self._db_migration_ran = True
 
             self._config_provider.set_airflow_config(
                 self._config_generator.config_template,
--- a/src/charm.py
+++ b/src/charm.py
@@ -9,7 +9,9 @@ import logging
 import charms.airflow_coordinator_k8s.v0.airflow_coordinator as airflow_coordinator
 import charms.data_platform_libs.v0.data_interfaces as data_interfaces_v0
 import ops
+from ops.pebble import LayerDict
 
+import command_executor
 import config_generator
 import constants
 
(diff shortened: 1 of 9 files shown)


src/charm.py
```diff
@@ -125,13 +170,49 @@ def _perform_checks(self) -> None:
...
         if not self.unit.is_leader():
             return
 
         try:
             self._perform_checks()
+            self._configure_pebble_layer()
+            self._write_airflow_config()
+
+            # Use StorageState to track if db migration has run before
+            if self._stored.db_migration_ran is False:
```
jananisenthilkumar291: I think if we never re-run `airflow db migrate` after the first success, the charm can incorrectly assume the database schema is still up to date even when something meaningful changes later (e.g., an Airflow upgrade introduces new migrations, or the Postgres relation changes endpoints/credentials and we effectively point at a different/restored database.

Instead, we could gate migrations so they run on “real” events such as on `pebble_ready` and `Postgres database_created/endpoints_changed`, and avoid running them during `update_status` to prevent log churn.

DnPlas: Yeah, that was my first approach (event driven), but using the reconciler makes it harder and prone to error as some of the events may be triggered and certain checks may fail, causing the migration to never run.

Having the ability to "switch" the state is beneficial because we can actually change it based on certain conditions, like the ones you mentioned. I did not add them at the moment, like the upgrade because we haven't decided how to handle it, the postgres credentials may be a good idea to add, but I'll hold the change until we decide our best route.

src/charm.py
```diff
@@ -33,13 +35,22 @@ def status(self):
 class AirflowCoordinatorK8SOperatorCharm(ops.CharmBase):
     """Charm the application."""
 
+    _stored = ops.StoredState()
```
shayancanonical: i was under the impression that we shouldn't be using stored state under any conditions anymore

should we instead add a peer relation and use the peer relation application databag?

src/charm.py
jananisenthilkumar291: From what I understand, most of airflow processes are idempotent including `airflow db migrate` as they use [Alembic migrations](https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html#:~:text=Whether%20to%20run%20alembic%20migrations%20during%20Airflow%20start%20up%2E), so they will not duplicate run if the schema/ set up is the same.
Hence, we could may be try to just avoid the `update_status` event and migrate on everything else, or we could avoid that check as well, as there'll be no harm in running `db migrate` multiple times (only concern is noisy logs).

DnPlas: It is true that the command is idempotent, BUT they also suggest to stop all services before running a db migration, thus ensuring this runs once (for now), should be safer than running on every event.

src/charm.py
DnPlas: After our discussion offline, I decided to go with peer relation instead since the state will be stored per application lifecycle and not per container, meaning we'll have more control over when we actually want to run the migration, and avoid running it every time the container restarts (which can happen often in k8s charms).

src/charm.py
jananisenthilkumar291: Hmm, That makes sense.

jananisenthilkumar291 (approved):
LGTM! Approved.

shayancanonical (approved):

PR #14, opened 2026-01-30
[WF-260] Implement support for custom airflow configuration

shayancanonical:
## Description

This PR adds:
1. config options (up to 16MiB) for non-sensitive custom airflow config (ini format)
2. juju user secret (up to 1MiB) for sensitve custom airflow config (ini format)

Validations include:
1. no overlap in config keys (`section.option`) between the two input configurations
2. blacklisted config keys defined -> not present in either input configurations

Configuration merge strategy: (base) template (overlay) sensitive (overlay) non-sensitive. overlay overrides values of existing keys

---
## Testing instructions

Please refer the [the changes introduced in the README](https://github.com/canonical/airflow-coordinator-k8s-operator/pull/14/changes#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R5) for a description of supported mechanisms to supply custom airflow configurations

To manually test, one can supply a disjoint ini file for the sensitive and non-sensitive custom configs (both with at least one value that overwrites a value in the auto generated config, and one that does not exist in the auto generated config). Relate coordinator with all required core charms, and ensure the workloads come up and the core charm workloads have the correct config files.

sensitive:
```
[api_auth]
jwt_secret = secret

[logging]
base_log_folder = /opt/airflow/sensitive_logs_directory # artificially simulating a sensitive key-value that is in the auto-generated config
```

non-sensitive
```
[core]
dag_folder = /opt/airflow/nonsensitive_dags_folder

[api]
enable_swagger_ui = False
```

after changes propagated to core charms, `juju ssh --container airflow-api-server airflow-api-server-k8s/0 cat /opt/airflow/airflow.cfg` to confirm appropriate values

Integration test implementation to follow shortly
---
## Checklist (all that apply)
- [ ] Documentation updated


jananisenthilkumar291 (changes requested):
Overall LGTM. Added minor nits.

src/charm.py
```diff
@@ -142,8 +144,29 @@ def _required_dependencies_exist(self) -> bool:
...
 
-    def _perform_checks(self) -> None:
+    def _perform_checks(self) -> None:  # noqa: C901
         """Checks to ensure the charm is able to generate and distribute configs."""
+        if self.config is None:
+            raise ExceptionWithStatusError(
+                constants.WAITING_FOR_CHARM_SETUP_MESSAGE, ops.WaitingStatus
+            )
+
+        if self.config.get(constants.SENSITIVE_CUSTOM_CONFIG):
```
jananisenthilkumar291: minor: can we assign secret_id once in a variable and use it as we're using it multiple times.

src/charm.py
```diff
@@ -142,8 +144,29 @@ def _required_dependencies_exist(self) -> bool:
...
+            try:
+                self.model.get_secret(
+                    id=self.config[constants.SENSITIVE_CUSTOM_CONFIG],
+                )
+            except ops.SecretNotFoundError as e:  # SecretNotFoundError is a subclass of ModelError
+                logger.error(e)
```
jananisenthilkumar291: minor: Can we add `exc_info=True` (or use `logger.exception`) to include stack trace and avoid writing sensitive info to logs.

src/config_generator.py
```diff
@@ -3,26 +3,132 @@
...
+]
+
 
 class AirflowConfigGenerator:
     """Encapsulate Airflow config generation logic."""
 
     def __init__(self, charm: ops.CharmBase):
         self._charm = charm
 
+        self._custom_config_parser, self._sensitive_custom_config_parser = None, None
```
jananisenthilkumar291: Can we have the `__init__` lighter and have a model to load custom config and call it here?

src/config_generator.py
```diff
@@ -3,26 +3,132 @@
...
+                    id=self._charm.config[constants.SENSITIVE_CUSTOM_CONFIG],
+                )
+
+                self._sensitive_custom_config_parser = configparser.ConfigParser()
+                self._sensitive_custom_config_parser.read_string(
+                    custom_config_secret.get_content(refresh=True)[
+                        constants.SENSITIVE_CUSTOM_CONFIG_SECRET_KEY
+                    ],
+                )
+            except (ops.model.SecretNotFoundError, ops.model.ModelError):
```
jananisenthilkumar291: Can we have a logger to surface this, if there's trouble loading sensitive custom config secret.

src/config_generator.py
```diff
@@ -3,26 +3,132 @@
...
+
+        if self._sensitive_custom_config_parser:
+            for section in self._sensitive_custom_config_parser.sections():
+                for option in self._sensitive_custom_config_parser.options(section):
+                    if not final_config_parser.has_section(section):
+                        final_config_parser.add_section(section)
 
-        return config_template
+                    final_config_parser.set(
+                        section, option, f"{{{{ {section}_{option}_secret_value }}}}"
```
jananisenthilkumar291: Out of curiosity: if a user defines a section or option containing a hyphen (e.g. my-section), Jinja will interpret that as a subtraction expression rather than a variable name. Are we guarding against this anywhere?

If not, should we consider storing sensitive values in a dict and referencing them via a lookup (e.g. `{{ sensitive["<key>"] }}`) to avoid invalid Jinja identifiers?

src/charm.py
```diff
@@ -142,8 +144,29 @@ def _required_dependencies_exist(self) -> bool:
             ]
         )
 
-    def _perform_checks(self) -> None:
+    def _perform_checks(self) -> None:  # noqa: C901
         """Checks to ensure the charm is able to generate and distribute configs."""
+        if self.config is None:
+            raise ExceptionWithStatusError(
+                constants.WAITING_FOR_CHARM_SETUP_MESSAGE, ops.WaitingStatus
+            )
```
DnPlas: I think we're trying to prevent the charm config from not being loaded initially, but is that even the case?

Please let me know if we are trying to do something else.

charmcraft.yaml
```diff
@@ -59,3 +59,12 @@ requires:
...
     interface: coordinator_peers
+
+config:
+  options:
+    sensitive_airflow_configuration_secret:
+      type: secret
+      description: "URI for secret that contains the user specified Airflow config with sensitive values"
+    custom_airflow_configuration:
+      type: string
+      description: "Custom Airflow config without sensitive values"
```
DnPlas: I think it'd be a good idea to provide some guidance on how these should be provided, perhaps an example can help clarify it as these point to multi key secret and a configuration file.

DnPlas (commented):
Thanks @shayancanonical, couple comments. Could you please add manual testing instructions?

src/charm.py
shayancanonical: i prefer not to in this particular case as `self.config` is already a class level variable and it is easier and cleaner to just access a key in the dictionary directly instead of introducing a new class level variable

src/charm.py
shayancanonical: moved to using `logger.exception` in ece934b5e2720dc6d14d7b0580453d686633c055

src/config_generator.py
shayancanonical: honestly, in my opinion, the current assignments of the config parsers is quite simple so far 

for custom config, we just access `charm.config[<config_key>]`
for sensitive custom config, a couple more steps which include getting the secret and getting the secret content

i dont feel good about adding an abstraction layer for this as it does not seem heavy enough to justify it

updated in ece934b5e2720dc6d14d7b0580453d686633c055: added a property in charm object to get the sensitive config secret. this hopefully makes it a bit more simple and less repetitive

src/config_generator.py
shayancanonical: the logger will surface this in `charm.py` in the `_perform_checks` method. this is why i dont believe it necessary to log errors in `config_generator` as well

src/config_generator.py
shayancanonical: great catch! addressed in ece934b5e2720dc6d14d7b0580453d686633c055 (using the suggestion to use a dict to wrap custom config secret values)

src/charm.py
shayancanonical: ah, i was under the impression that the `start` event can be run before `config-changed` (mixing it up with `install`). this was supposed to be a guard for this use case since we use a reconcile pattern

after confirming that configs will always be loaded for all observed events, removed this condition/check in 305882f9cc94c35db74ab7ff9bf8dda07b2daa71

charmcraft.yaml
shayancanonical: added some preliminary user guidance in the readme: 305882f9cc94c35db74ab7ff9bf8dda07b2daa71

src/config_generator.py
jananisenthilkumar291: Must've Missed that. Thanks Shayan

charmcraft.yaml
DnPlas: That is also okay, but since this goes to charmhub.io, I'd strongly suggest to have a slightly more descriptive config.

shayancanonical: I agree with the need to make the doc a bit more descriptive and improve structure

I was under the impression that GH readmes dont stream to charmhub.io - we'd be required to create discourse.charmhub posts to populate charmhub docs. When this happens, I think the README content related to custom config should be deleted

DnPlas: I mean make the config more descriptive, not the doc. Similar to https://github.com/canonical/istio-operators/blob/main/charms/istio-pilot/config.yaml#L41

src/config_generator.py
```diff
@@ -3,26 +3,130 @@
...
     """Encapsulate Airflow config generation logic."""
 
     def __init__(self, charm: ops.CharmBase):
         self._charm = charm
 
+        self._custom_config_parser, self._sensitive_custom_config_parser = None, None
+
+        if self._charm.config.get(constants.CUSTOM_CONFIG):
+            self._custom_config_parser = configparser.ConfigParser()
+            self._custom_config_parser.read_string(self._charm.config[constants.CUSTOM_CONFIG])
```
DnPlas: This may fail if there is a duplication error (and there could even be more errors AFAIK), but since there is no error handling, the coordinator charm will go into error status.

```
Unit              Workload  Agent  Address     Ports  Message             
api-server/0*     active    idle   10.1.0.101                                                    
coordinator/0*    error     idle   10.1.0.225         hook failed: "config-changed"    

# From juju debug-logs

  File "/var/lib/juju/agents/unit-coordinator-0/charm/venv/lib/python3.12/site-packages/ops/_main.py", line 313, in __init__
    self.charm = self._charm_class(self.framework)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/lib/juju/agents/unit-coordinator-0/charm/src/charm.py", line 43, in __init__
    self._config_generator = config_generator.AirflowConfigGenerator(self)
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/lib/juju/agents/unit-coordinator-0/charm/src/config_generator.py", line 36, in __init__
    self._custom_config_parser.read_string(self._charm.config[constants.CUSTOM_CONFIG])
  File "/usr/lib/python3.12/configparser.py", line 710, in read_string
    self.read_file(sfile, source)
  File "/usr/lib/python3.12/configparser.py", line 705, in read_file
    self._read(f, source)
  File "/usr/lib/python3.12/configparser.py", line 1075, in _read
    raise DuplicateOptionError(sectname, optname,
configparser.DuplicateOptionError: While reading from '<string>' [line  3]: option 'dag_folder' in section 'core' already exists

```

Nothing seems to resolve the error, not even passing a different configuration.

Let's make sure we provide a better error handling and unit status.

src/config_generator.py
```diff
@@ -3,26 +3,130 @@
...
+import constants
+
 logger = logging.getLogger(__name__)
 
+BlacklistConfigKey = collections.namedtuple("BlacklistConfigKey", ["section", "option"])
+
+BLACKLIST_CUSTOM_CONFIG_KEYS = [
+    BlacklistConfigKey(section="core", option="executor"),
+    BlacklistConfigKey(section="database", option="sql_alchemy_conn"),
+]
```
DnPlas, line 23: If we allow people to change the dags folder, we need to ensure dags are actually stored there - IMHO, maybe it's not worth it as we are more or less in control of that(?)

README.md
```diff
@@ -1,3 +1,48 @@
...
+
+[logging]
+base_log_folder = /opt/airflow/logs
+```
+
+For configuration values that are sensitive, we provide a way to specify a
+juju user secret with the config file (up to 1MiB):
+
+```
+$ juju add-secret custom_config_secret sensitive_airflow_configuration_secret#file=/tmp/sensitive_config.ini
```
DnPlas: juju will not allow keys to use underscores, please s/sensitive_airflow_configuration_secret/sensitive-airflow-configuration-secret

```
juju add-secret my_secret sensitive_airflow_configuration_secret#file=sensible.ini
ERROR key "sensitive_airflow_configuration_secret" not valid
```

README.md
```diff
@@ -1,3 +1,48 @@
...
+$ juju config sensitive_airflow_configuration_secret="<URI of custom_config_secret>"
```
DnPlas:
```suggestion
$ juju config airflow-coordinator-k8s sensitive_airflow_configuration_secret="<URI of custom_config_secret>"
```

README.md
DnPlas: We also need to grant the coordinator this secret.

DnPlas: I believe the naming will cause an issue:

```
unit-coordinator-0: 11:28:50 ERROR unit.coordinator/0.juju-log Uncaught exception while in charm code:   
Traceback (most recent call last):                                                                       
  File "/var/lib/juju/agents/unit-coordinator-0/charm/src/charm.py", line 291, in <module>               
    ops.main(AirflowCoordinatorK8SOperatorCharm)
  File "/var/lib/juju/agents/unit-coordinator-0/charm/venv/lib/python3.12/site-packages/ops/__init__.py", line 356, in __call__
    return _main.main(charm_class=charm_class, use_juju_for_storage=use_juju_for_storage)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/lib/juju/agents/unit-coordinator-0/charm/venv/lib/python3.12/site-packages/ops/_main.py", line 494, in main
    manager = _Manager(charm_class, use_juju_for_storage=use_juju_for_storage)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/lib/juju/agents/unit-coordinator-0/charm/venv/lib/python3.12/site-packages/ops/_main.py", line 313, in __init__
    self.charm = self._charm_class(self.framework)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/lib/juju/agents/unit-coordinator-0/charm/src/charm.py", line 43, in __init__
    self._config_generator = config_generator.AirflowConfigGenerator(self)
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/lib/juju/agents/unit-coordinator-0/charm/src/config_generator.py", line 42, in __init__
    custom_config_secret.get_content(refresh=True)[
KeyError: 'sensitive-custom-airflow-configuration'
```

src/constants.py
```diff
@@ -13,20 +13,25 @@
...
-    "Waiting for airflow database to be created"
-)
-WAITING_FOR_DATABASE_CONNECTION_MESSAGE = (
-    "Waiting for database connection info from postgres"
+UNAUTHORIZED_ACCESS_TO_SECRET_MESSAGE = (
+    "Charm is unauthorized to access sensitive custom config secret"
 )
+CUSTOM_CONFIG_SECRET_NOT_FOUND = "Sensitive custom config secret not found"
+CUSTOM_CONFIG_OVERLAP_MESSAGE = "Sensitive and non-sensitive custom configs have overlap"
+CUSTOM_CONFIG_HAS_BLACKLIST_KEY = "Sensitive or non-sensitive custom configs have blacklisted keys"
```
DnPlas: (suggestion, non-blocking) let's use blocklist or denylist(?)

DnPlas: Also, it may be a good idea to show the blocklisted keys in the logs, so this msg could be something like:

```
"Sensitive or non-sensitive custom configs have bocked keys, see logs for more details."
```

DnPlas (commented):
Thanks @shayancanonical, some comments.

charmcraft.yaml
shayancanonical: Updated sensitive config secret description based on the istio reference: 66258caab14fe7904def58ddb7f2bdf8b6f45663

src/config_generator.py
shayancanonical: now handling any `configparser.Error`: 66258caab14fe7904def58ddb7f2bdf8b6f45663

the charm leader unit goes into `BlockedStatus` if there are config parser errors

src/config_generator.py
shayancanonical: I agree - we would have nothing to enforce or validate at this time. Until we allow `LocalDagBundles`, the dags in the provided `dags_folder` will come exclusively from git/s3. When we add support for `LocalDagBundle`, we can add a validation - or we would need to address the local dag bundles not being present in the configured directory

README.md
shayancanonical: updated the README (changed secret name to use - instead of _): 66258caab14fe7904def58ddb7f2bdf8b6f45663

README.md
shayancanonical: updated in 66258caab14fe7904def58ddb7f2bdf8b6f45663

README.md
shayancanonical: yep! added the grant-secret command: 66258caab14fe7904def58ddb7f2bdf8b6f45663

src/constants.py
shayancanonical: updated message in66258caab14fe7904def58ddb7f2bdf8b6f45663

imo, i like the use of blacklist as a variable descriptor - it's consistent with blacklist/whitelist semantics. please let me know if this is still confusing, happy to rename

DnPlas (dismissed):
Thanks @shayancanonical, looks good overall, I'll leave that one thread about whitelist/blacklist open, but feel free to close it if there'll be no changes.

src/constants.py
shayancanonical: updated blacklist -> denylist in https://github.com/canonical/airflow-coordinator-k8s-operator/pull/14/commits/61427b97c67222c53da53c9e3aa7df0bf6035092

DnPlas (approved):

jananisenthilkumar291 (approved):
LGTM!

PR #21, opened 2026-02-27
[WF-293] Update coordinator lib skip unchanged airflow.cfg writes

nikita-jare:
## Description

This change updates the Airflow Coordinator shared charm library to detect whether rendered `airflow.cfg` content actually changed before writing to the workload container.

Changes included:
- `write_airflow_config()` now compares existing file content with rendered content
- file write is skipped when content is unchanged
- function now returns `bool`:
  - `True` when content changed and file was written
  - `False` when content was unchanged and write was skipped
- `AirflowCoordinatorRequires.write_airflow_config()` now returns that `bool` to callers
- `LIBPATCH` is bumped to publish the updated library revision

Related to canonical/airflow-core-operators#19

---
## Testing instructions

1. Build & deploy coordinator + all four core charms.
2. Ensure all required integrations are present and model is active.
3. Capture baseline service processes:
```
juju ssh --container airflow-api-server airflow-api-server-k8s/0 "pgrep -fa 'airflow api_server'"
juju ssh --container airflow-scheduler airflow-scheduler-k8s/0 "pgrep -fa 'airflow scheduler'"
juju ssh --container airflow-dag-processor airflow-dag-processor-k8s/0 "pgrep -fa 'airflow dag-processor'"
juju ssh --container airflow-triggerer airflow-triggerer-k8s/0 "pgrep -fa 'airflow triggerer'"
```
4. Trigger config change from coordinator:
Edit coordinator airflow_config.j2 (example:change  load_examples=False to True)

5. Repack and refresh coordinator:
```
juju refresh airflow-coordinator-k8s --path <local-airflow-coordinator-k8s.charm> --resource airflow-coordinator-image=ubuntu/airflow:3.1-24.04_edge
```
Verify change applied and workloads restarted.
Expected:
- Updated config value is present across core charms.
- Core workload processes changed from baseline after real config change.

7. Verify no-op refresh does not restart
Repack/refresh coordinator again without changing template.
Re-check process IDs and service state.
Expected:
- No new config write for unchanged content.
- Core workload processes remain stable on no-op refresh

---
## Checklist (all that apply)
- [X] Unit tests added or updated
- [ ] Documentation updated

### DIFF
--- a/lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
+++ b/lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
@@ -172,12 +172,46 @@ LIBPATCH = 3
 logger = logging.getLogger(__name__)
 
 
-def write_airflow_config(
+def airflow_config_needs_update(
     container: ops.Container,
     config_path: str,
     config_template: str,
     sensitive_data: dict[str, str],
 ) -> bool:
+    """Check whether the rendered Airflow config differs from the file currently on disk.
+
+    Args:
+        container: The Pebble container to check the config in.
+        config_path: Path to the config file to compare against.
+        config_template: The Jinja2 template string for the Airflow config.
+        sensitive_data: Dictionary of sensitive values to render in the template.
+
+    Returns:
+        bool: True if the config needs to be written (content differs or file absent), else False.
+
+    Raises:
+        RuntimeError: If unable to connect to the container.
+    """
+    if not container.can_connect():
+        raise RuntimeError("Cannot connect to workload container")
+
+    try:
+        config = jinja2.Environment().from_string(config_template).render(**sensitive_data)
+
+        if container.exists(config_path) and container.pull(config_path).read() == config:
+            return False
+
+        return True
+    except Exception as e:
+        raise RuntimeError(f"Failed to check Airflow config: {e}") from e
+
+
+def write_airflow_config(
+    container: ops.Container,
+    config_path: str,
+    config_template: str,
+    sensitive_data: dict[str, str],
+) -> None:
     """Render and write the Airflow config to a container.
 
     This utility function can be used by both the coordinator charm (for db migrations)
@@ -189,9 +223,6 @@ def write_airflow_config(
         config_template: The Jinja2 template string for the Airflow config.
         sensitive_data: Dictionary of sensitive values to render in the template.
 
-    Returns:
-        bool: True if the file content changed and was written, else False.
-
     Raises:
         RuntimeError: If unable to connect to the container or write the config.
     """
@@ -201,10 +232,6 @@ def write_airflow_config(
     try:
         config = jinja2.Template(config_template).render(**sensitive_data)
 
-        if container.exists(config_path) and container.pull(config_path).read() == config:
-            logger.info("Airflow config at %s is unchanged; skipping write", config_path)
-            return False
-
         container.push(
             config_path,
             config,
@@ -213,7 +240,6 @@ def write_airflow_config(
             make_dirs=True,
         )
         logger.info("Successfully wrote Airflow config to %s", config_path)
-        return True
     except Exception as e:
         raise RuntimeError(f"Failed to write Airflow config: {e}") from e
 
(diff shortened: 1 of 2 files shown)


tests/unit/test_airflow_coordinator.py
```diff
@@ -428,6 +428,90 @@ def test_write_airflow_config(
             assert config_file_path.is_file()
             assert config_file_path.read_text(encoding="utf-8") == "test-config: s3cret"
             assert config_file_path.stat().st_mode & 0o777 == 0o644
+            assert config_changed
+
+    def test_write_airflow_config_returns_false_when_unchanged(self):
+        """Skip write and return False when rendered config is unchanged."""
```
jananisenthilkumar291: Can we mock container using `unittest.mock` and use `@pytest.mark.parametrize` to test both the scenarios.

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
```diff
@@ -154,6 +157,12 @@ def write_airflow_config(
     try:
         config = jinja2.Environment().from_string(config_template).render(**sensitive_data)
 
+        if container.exists(config_path):
```
jananisenthilkumar291: I think we can collapse this into a single if block. `current_config` is not being used anywhere else, so this assignment can be remove as well.
```
if container.exists(config_path) and container.pull(config_path).read() == config:
     logger.info(f"Airflow config at {config_path} is unchanged; skipping write")
     return False
```

jananisenthilkumar291 (changes requested):
LGTM! Will approve after outstanding comments are resolved.

tests/unit/test_airflow_coordinator.py
nikita-jare: Added in latest [commit](https://github.com/canonical/airflow-coordinator-k8s-operator/pull/21/commits/d77391361fb32e8ca2f396c5bb22ad24ae409628)

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
```diff
@@ -133,7 +133,7 @@ def write_airflow_config(
     config_path: str,
     config_template: str,
     sensitive_data: dict[str, str],
-) -> None:
+) -> bool:
```
shayancanonical: nit: i prefer to add another method that will return a bool to indicate whether the airflow config will be updated (e.g. a property called `airflow_config_should_be_updated`). this way, the `write_airflow_config` is idempotent such that it will always write the file (we dont ever call the method and expect it to not write the file)

shayancanonical (changes requested):

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
```diff
@@ -121,13 +121,47 @@ def reconcile(self, event) -> None:
...
 # to 0 if you are raising the major API version
-LIBPATCH = 2
+LIBPATCH = 3
 
 # TODO: add your code here! Happy coding!
 
 logger = logging.getLogger(__name__)
 
 
+def airflow_config_needs_update(
```
shayancanonical: is this helper function used anywhere?

tests/unit/test_airflow_coordinator.py
```diff
@@ -429,6 +430,88 @@ def test_write_airflow_config(
...
+        """Return True when config differs from disk or file is absent, False when unchanged."""
+        pulled_file = unittest.mock.Mock()
+        pulled_file.read.return_value = existing_content
+
+        container = unittest.mock.Mock()
+        container.can_connect.return_value = True
+        container.exists.return_value = file_exists
+        container.pull.return_value = pulled_file
+
+        needs_update = airflow_coordinator.airflow_config_needs_update(
```
shayancanonical: nit: we should call `manager.charm.requirer.airflow_config_needs_update`. to gain access to manager, we can use the context manager (like in the previous test) to invoke `update_status`

shayancanonical (commented):
looks great! a couple of comments - ready to approve after they are addressed

tests/unit/test_airflow_coordinator.py
nikita-jare: Removed the parametrized test since the 3 scenarios are already covered by the 3 class-level tests

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
nikita-jare: removed the helper function

shayancanonical (approved):

jananisenthilkumar291 (approved):

PR #22, opened 2026-02-27
refactor: enable more generic configuration passing using the airflow_coordinator library

DnPlas:
## Description
This commit adds a new abstraction for enabling more generic configuration passing without the need of the requirer sending over its component name, container, etc. There are now two classes:
* AirflowCoordinatorCoreRequires - responsible for all the sharing, validation, and writing of the configuration on the Airflow core components (this is the previous behaviour)
* AirflowCoordinatorRequires - responsible for sharing and writing data, but with less validation, and with less strict requirements from the requirer. This becomes a parent class of AirflowCoordinatorCoreRequires.

---
## Testing instructions

NA

This is the first PR on a series of multiple PRs to enable the Kubernetes Executor functionality and charm.

---
## Checklist (all that apply)
- [ ] Documentation updated

### DIFF
--- a/lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
+++ b/lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
@@ -165,7 +165,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 2
+LIBPATCH = 3
 
 # TODO: add your code here! Happy coding!
 
--- a/lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
+++ b/lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
@@ -824,7 +824,6 @@ class AirflowCoordinatorRequires(ops.Object):
         Args:
             config_path: the path where the configuration file will be saved.
         """
-
         provider_content = self.provider_content
         config = jinja2.Template(provider_content.config_template).render(
             **json.loads(provider_content.sensitive_data)
--- a/lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
+++ b/lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
@@ -812,7 +812,7 @@ class AirflowCoordinatorRequires(ops.Object):
         This check ensures the coordinator has shared relevant config data
         in the relation and that the relation exists.
         """
-        content = self._requirer_handler.provider_content
+        content = self.provider_content
         return bool(content and content.config_template and content.sensitive_data)
 
     def write_airflow_config(self, config_path: str) -> None:
@@ -825,7 +825,7 @@ class AirflowCoordinatorRequires(ops.Object):
             config_path: the path where the configuration file will be saved.
         """
 
-        provider_content = self._requirer_handler.provider_content
+        provider_content = self.provider_content
         config = jinja2.Template(provider_content.config_template).render(
             **json.loads(provider_content.sensitive_data)
         )
--- a/lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
+++ b/lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
@@ -11,9 +11,50 @@ retrieval of fields in pydantic models in the databag if plaintext, and in a
 juju secret if sensitive. We expand upon the implementation approach established
 in data_interfaces v1.
 
-### Requirer Charm
+### Requirer Charms
 
-The following presents an example usage of the AirflowCoordinatorRequires class:
+Two requirer classes are provided depending on the nature of the charm.
+
+**`AirflowCoordinatorRequires`** is the base class for charms that consume the
+coordinator relation but do not run an Airflow workload in a sidecar container
+(e.g. the coordinator charm itself running DB migrations). It writes the rendered
+Airflow config directly to the local filesystem via `pathlib`.
+
+```python
+import charms.airflow_coordinator_k8s.v0.airflow_coordinator as airflow_coordinator
+
+class AirflowRequirerCharm(ops.CharmBase):
+    def __init__(self, *args) -> None:
+        super().__init__(*args)
+
+        self.requirer = airflow_coordinator.AirflowCoordinatorRequires(
+            self,
+            "airflow-coordinator", # relation endpoint
+            callback=self.reconcile,
+        )
+
+    def reconcile(self, event) -> None:
+        # Determine current state of charm, what it should be, and how to get there
+```
+
+`AirflowCoordinatorRequires` surfaces the following:
+1. `provider_content`: the raw data shared by the coordinator charm
+2. `can_write_airflow_config`: True when the coordinator has shared a config
+template and sensitive data
+3. `write_airflow_config(filepath)`: renders and writes the Airflow config to
+a local filesystem path
+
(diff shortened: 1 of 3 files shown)


lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
```diff
@@ -733,8 +774,85 @@ def __init__(
...
+    def write_airflow_config(self, config_path: str) -> None:
+        """Render the Airflow config and write it to a local filesystem path.
+
+        Writes to a plain filesystem path rather than a pebble workload container,
+        making it suitable for charms without a sidecar container.
+
+        Args:
+            config_path: the path where the configuration file will be saved.
+        """
+        import pathlib
```
shayancanonical: any reason this package is imported inline? can we move it to a global file import instead?

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
```diff
@@ -733,8 +774,85 @@ def __init__(
...
+
+        provider_content = self._requirer_handler.provider_content
```
shayancanonical: nit: we should update to use `self.provider_content`, even within the class. this way, any updates to how the provider content is created and/or parsed will be localized to the property

tests/unit/test_airflow_coordinator.py
```diff
@@ -277,38 +326,188 @@ def generate_coordinator_state(
...
+            base_requires_context.on.relation_changed(
+                application_airflow_coordinator_relation
+            ),
+            base_application_state,
+        ) as manager:
+            manager.run()
+            manager.charm.requirer.write_airflow_config(config_path)
+
+        assert pathlib.Path(config_path).exists()
+        assert pathlib.Path(config_path).read_text() == "test-config: s3cret"
```
shayancanonical: nit: running on local machine will result in build up of tmp files (at the mercy of the system's cleanup)

can be slightly improved by using `tempfile.NamedTemporaryFile` in a context manager and using the tempfile's name as the config_path. then when the context manager is naturally exited, the temp file is auto cleaned

tests/unit/test_airflow_coordinator.py
```diff
@@ -277,38 +326,188 @@ def generate_coordinator_state(
...
+
+    def test_write_airflow_config_creates_parent_directories(
+        self,
+        base_requires_context,
+        base_application_state,
+        application_airflow_coordinator_relation,
+        tmp_path,
+    ):
+        config_path = str(tmp_path / "deeply" / "nested" / "dir" / "airflow.cfg")
```
shayancanonical: recommend recursively cleaning the `deeply/` directory in case of successful test runs

shayancanonical (commented):
looks great

im very curious as to how the config_template and executor_config work together. i am assuming that the config_template is expected to contain partial templates to be shared with the k8s executor charm. then the k8s executor charm updates this template and sends it back to the coordinator to share with the downstream core charms. i will confirm this assumption

if this is the case, a cleaner alternative may be to add a new field to the pydantic model and keep the 2 values separate at all times. to then also update the core charms to always use the value of the `executor_config` (or equivalent) if present, and if not, then default to `config_template`. this may make it more traceable.

thoughts?

DnPlas:
> looks great
> 
> im very curious as to how the config_template and executor_config work together. i am assuming that the config_template is expected to contain partial templates to be shared with the k8s executor charm. then the k8s executor charm updates this template and sends it back to the coordinator to share with the downstream core charms. i will confirm this assumption
> 
> if this is the case, a cleaner alternative may be to add a new field to the pydantic model and keep the 2 values separate at all times. to then also update the core charms to always use the value of the `executor_config` (or equivalent) if present, and if not, then default to `config_template`. this may make it more traceable.
> 
> thoughts?

Close, but not quite. The coordinator generates the config_template itself (from a local Jinja2 file), not via a round-trip through the k8s executor charm. The executor_config parameter is an alternative input format: a plain dict of section → key/value pairs (e.g. {"executor": {"default_queue": "kubernetes"}}), which the provider side immediately JSON-serializes and stores into config_template.

If you check https://github.com/canonical/airflow-coordinator-k8s-operator/pull/23 you will see how this is implemented. I remember keeping in mind that we wanted to keep these two values separate form the beginning. Please let me know if we want to go with the pydantic model or the current approach. I'm open to either.

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
DnPlas: Fixed in [60517db](https://github.com/canonical/airflow-coordinator-k8s-operator/commit/60517db2ffc2a8b7116314b0faaa32a15e0c56fe)

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
DnPlas: Fixed in [ccf293e](https://github.com/canonical/airflow-coordinator-k8s-operator/commit/ccf293e857f559f2560395066fc262d1164214b9)

tests/unit/test_airflow_coordinator.py
DnPlas: Fixed in [ccf293e](https://github.com/canonical/airflow-coordinator-k8s-operator/commit/ccf293e857f559f2560395066fc262d1164214b9)

tests/unit/test_airflow_coordinator.py
DnPlas: Fixed in [ccf293e](https://github.com/canonical/airflow-coordinator-k8s-operator/commit/ccf293e857f559f2560395066fc262d1164214b9)

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
```diff
@@ -1000,20 +1115,27 @@ def set_validation_errors(self) -> None:
...
+            k8s_executor_pod_spec_template: (optional) K8s executor pod spec template.
             sensitive_data: sensitive data to render config of k8s executor pod
-                spec jinja templates with
+                spec jinja templates with.
+            executor_config: Airflow config as a dict of section -> key/value pairs
+                (e.g. {"executor": {"default_queue": "kubernetes"}}). Serialised to
+                JSON and stored in the relation databag. Takes precedence over
+                config_template when both are provided.
         """
+        template = json.dumps(executor_config) if executor_config is not None else config_template
```
shayancanonical: to confirm, this will be further changed in subsequent PRs right? in particular, the config_generator.py will get involved in how to overlay the render template and the additional config key-values coming from the k8s executor charm?

shayancanonical (commented):
one question to confirm a small point, and a nit. lgtm otherwise - will approve after those two are considered

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
DnPlas: I'll actually just remove it from this PR, we can leave the config_generator decide how to proceed (just append the extra configuration).

DnPlas: Fixed in 7d8bea7

shayancanonical (changes requested):
Note: you will need to bump the libpatch to 3

Want to select `Approve` instead of `Request changes`, but want to make sure that the charm lib gets updated to keep publish lib job happy. Will approve right after update of libpatch

shayancanonical (approved):

PR #23, opened 2026-02-28
feat: integrate with the executor interface

DnPlas:
## Description
Integrate the coordinator with the Kubernetes executor charm via the airflow-executor-config relation,
enabling dynamic executor configuration and pod spec distribution to core Airflow charms.

Key changes:
- Add airflow-executor-config relation endpoint to receive executor configuration
- Refactor config template to return raw Jinja2 (no pre-rendering), with API server and executor
values merged as structured {section: {key: value}} extra config via config_template_with_extra_config
- Config merge that replaces existing keys, inserts new keys under existing sections, and appends new sections using configparser
- Distribute the executor's pod spec template (kubernetes-executor-pod-spec) to core charms
- Rename sensitive config key from sql_alchemy_connection_string to database__sql_alchemy_conn to match Airflow's <section>__<key> convention

Library changes (airflow_coordinator v0, LIBPATCH 4 -> 5):
- Widen config_template type to str | dict | None to handle get_data()'s automatic JSON
deserialization of structured config from the executor
- Make secret_sensitive_data type SecretString | None so validation doesn't fail when the executor
relation doesn't provide secrets
- Remove truthiness guard on kubernetes_executor_pod_spec in update_content so None clears stale pod
specs after relation removal
- Log ValidationError in provider_content instead of silently swallowing it

---
## Testing instructions

1. charmcraft pack the coordinator                                                                    
2. Deploy with postgresql-k8s, airflow api server, scheduler, triggerer, and the Kubernetes executor charm from https://github.com/canonical/airflow-kubernetes-executor-k8s-operator/pull/2
3. Integrate coordinator and executor through airflow-coordinator and airflow-executor-config relations. 
4. Check that core charms are receiving k8s executor config in their airflow.cfg `kubectl exec -ti -nairflow scheduler-0 -c airflow-scheduler -- cat /opt/airflow/airflow.cfg`
5. Check relation data bag for k8s pod spec (juju show-unit coordinator should suffice)                                                             
6. Remove the executor integration: verify the coordinator falls back to LocalExecutor and clears the pod spec from core charm databags                                                                    
7. Re-add the executor integration: verify KubernetesExecutor config is restored 

---
## Checklist (all that apply)
- [ ] Documentation updated

### DIFF
--- a/src/config_generator.py
+++ b/src/config_generator.py
@@ -3,9 +3,10 @@
 
 """Charm support for the Airflow config generation."""
 
+import configparser
+import io
 import logging
 import pathlib
-import re
 
 import ops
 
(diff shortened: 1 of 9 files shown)


src/config_generator.py
```diff
@@ -53,9 +47,85 @@ def _sql_alchemy_connection_string(self) -> str:
...
+        # changes to the library and the core charms.
+        result = self.config_template
+        if not extra_config:
+            return result
+
+        new_sections = []
+        for section, keys in extra_config.items():
+            body = self._section_body(result, section)
+            if body is not None:
+                existing_keys = set(re.findall(r"^(\w+)\s*=", body, re.MULTILINE))
```
shayancanonical: i believe the contents of this method can be significantly simplified with `configparser.RawConfigParser` which will avoid parsing `{{ }}` and `{% %}`

this means that as long as the k8s config key-values are not in conflict with the templated key-values in `airflow_config.j2`, this can be resolved quite nicely

if there are conflicts, we can also parse much cleaner (iterating over `p.sections()` and `p.options()` seems much more robust than using regex)

```
>>> p.read_string('[a]\na = {{ a }}\n{% if b %}b = {{ b }}{% endif %}\nc = "c"\n')
>>> p.sections()
['a']
>>> p.options("a")
['a', '{% if b %}b', 'c', 'b']
>>> p.get("a", "{% if b %}b")
'{{ b }}{% endif %}'
```

DnPlas: Fixed in fd8cae6

tests/unit/test_charm.py
```diff
@@ -352,6 +354,128 @@ def test_db_migration_failure(context, state, mock_command_executor, workload_co
...
+            AirflowCoordinatorK8SOperatorCharm, "_kubernetes_executor_config",
+            new_callable=unittest.mock.PropertyMock(return_value=MOCK_KUBERNETES_EXECUTOR_CONFIG),
+        ):
+            state_out = context.run(context.on.start(), state)
+
+        assert state_out.unit_status == ops.ActiveStatus()
+
+        for relation in state_out.get_relations("airflow-coordinator"):
+            config_template = relation.local_app_data.get("config-template", "")
+            assert "[kubernetes_executor]" in config_template
```
shayancanonical: (nit, non-blocking, feel free to ignore): should we also check to confirm that `core.executor = kuberenetes_executor`? to ensure that proper overwrites of existing keys also happens in the final config ini file sent to the core charms?

if we do that, then we can merge this test with `test_kubernetes_executor_config_sets_executor_in_core_section` into a single test with the same assertions

shayancanonical (dismissed):
looks really great! one minor nit about possibility to combine 2 of the unit tests. feel free to ignore, but also feel free to ping for re-approval if a new commit is pushed

tests/unit/test_charm.py
DnPlas: Addressed in 0dcb9e3

shayancanonical (approved):

PR #27, opened 2026-03-04
[WF-303] Removed dags and logs config from config template

jananisenthilkumar291:
## Description

The current coordinator configuration template sets:

```
dags_folder = dags
base_log_folder = logs
```
Since these are absolute paths, they map to /dags instead of $AIRFLOW_HOME/dags which is where airflow typically looks for dags.

Since as per [documented](https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html#:~:text=Default%3A-,%7BAIRFLOW_HOME%7D/dags,-Environment%20Variable%3A), the default values to these config options are already consistent with the expected, these configs can be entirely eliminated from the config template. However, updating to absolute paths /opt/airflow/dags and /opt/airflow/logs is also an option.

---

---
## Checklist (all that apply)
- [ ] Documentation updated

### DIFF
--- a/justfile
+++ b/justfile
@@ -19,7 +19,7 @@ format: (clean-mock-charm-libs)
 	uv tool run tox -e format
 
 # Run integration tests
-integration debug="": (clean-mock-charm-libs)
+integration debug="": (clean)
 	#!/usr/bin/bash
 	charmcraft pack
 
@@ -39,7 +39,7 @@ unit:
 
 # Clean up test environment
 clean: (clean-mock-charm-libs)
-	juju destroy-model --force --destroy-storage --no-prompt test
+	juju destroy-model --force --destroy-storage --no-prompt test || true
 
 # Get system state for debugging
 get-system-state:
--- a/src/command_executor.py
+++ b/src/command_executor.py
@@ -14,9 +14,7 @@ logger = logging.getLogger(__name__)
 class CommandExecutionError(Exception):
     """Exception raised when command execution fails."""
 
-    def __init__(
-        self, message: str, return_code: int | None = None, stderr: str | None = None
-    ):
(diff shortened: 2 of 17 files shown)


justfile
```diff
@@ -39,7 +39,7 @@ unit:
 
 # Clean up test environment
 clean: (clean-mock-charm-libs)
-	juju destroy-model --force --destroy-storage --no-prompt test
+	juju destroy-model --force --destroy-storage --no-prompt test || true
```
DnPlas, line 42: Is it hanging on this step, or why are we adding this?

src/command_executor.py
```diff
@@ -59,8 +59,8 @@ def run_db_migrate(self) -> CommandExecutionResult:
             process = self._container.exec(
                 ["airflow", "db", "migrate"],
                 environment={
-                    "AIRFLOW_HOME": "/",
-                    "AIRFLOW__CORE__DAGS_FOLDER": "/dags",
+                    "AIRFLOW_HOME": "/opt/airflow",
+                    "AIRFLOW__CORE__DAGS_FOLDER": "/opt/airflow/dags",
```
DnPlas: I think this line is not required at all, the AIRFLOW HOME we could ditch by either passing the env var directly to the pebble layer or keeping it here, I'm okay either way.
```suggestion

```

DnPlas (commented):
Thanks @jananisenthilkumar291, two tiny little comments, other than that LGTM

justfile
jananisenthilkumar291: No. It doesn't hang. This is to ensure "idempotency" while running `just integration` locally as it invokes `just clean` which would otherwise fail if model is not present. And if the model is present, it cleans up the model, clears mock charm libs and proceeds to execute tests.

shayancanonical (approved):

PR #31, opened 2026-03-11
[WF-291] Added secret info in config

jananisenthilkumar291:
## Description

Related Issue:
[Link](https://github.com/canonical/airflow-core-operators/issues/17)

Added configs:
- api.secret_key -> CSRF Protection,
- api_auth.jwt_secret -> related to JWT token validity

---

---
## Checklist (all that apply)
- [ ] Documentation updated


src/charm.py
```diff
@@ -128,6 +130,51 @@ def _peer_relation(self) -> ops.Relation:
...
+
+        Must be identical across all API server units in HA mode.
+        Used for CSRF token generation, session cookie signing, and
+        Celery executor message signing. If inconsistent, CSRF validation
+        and session verification will fail across units.
+        """
+        secret = self._peer_relation.data[self.app].get("secret_key")
+        if not secret:
+            secret = secrets.token_hex(32)
+            self._peer_relation.data[self.app]["secret_key"] = secret
```
shayancanonical: i prefer if we did not store the keys in the peer relation databag in plaintext. rather, the charm can create a juju application secret (backed by the peer relation) that contains these values. the secret id can then be stored in the peer relation to retrieve values at runtime

shayancanonical (changes requested):

src/charm.py
jananisenthilkumar291: Good Point! Here's an update [098b8b9](https://github.com/canonical/airflow-coordinator-k8s-operator/pull/31/commits/098b8b9e6b413529ab3541d763b7b9ebd0c8ac6f)

src/charm.py
```diff
@@ -128,6 +131,39 @@ def _peer_relation(self) -> ops.Relation:
             )
         return peer_relation
 
+    @staticmethod
+    def _generate_runtime_secret_content(peer_data: Mapping[str, str]) -> dict[str, str]:
+        """Generate runtime secret values, preserving any legacy peer data values."""
+        return {
+            "secret-key": peer_data.get("secret_key") or secrets.token_hex(32),
```
shayancanonical: nit: simplification: `"secret-key": peer_data.get("secret_key", secrets.token_hex(32))`

can use the default value of `dict.get()` instead of using `or` (condition structure)

src/charm.py
```diff
@@ -128,6 +131,39 @@ def _peer_relation(self) -> ops.Relation:
             )
         return peer_relation
 
+    @staticmethod
```
shayancanonical: any particular reason to use a staticmethod? will it ever be called outside of the charm object's context?
suggestion: we should make this a normal method instead

suggestion: if this is a method instead, then a class-level property `peer_app_data` can be implemented to access the application databag of the peer relation. this way, no args will need to be passed into the method

src/charm.py
```diff
@@ -128,6 +131,39 @@ def _peer_relation(self) -> ops.Relation:
...
+        exist in the peer databag they are migrated into the secret and removed.
+        """
+        peer_data = self._peer_relation.data[self.app]
+
+        secret_id = peer_data.get(constants.AIRFLOW_RUNTIME_SECRET_ID_FIELD)
+        if secret_id:
+            return self.model.get_secret(id=secret_id)
+
+        try:
+            runtime_secret = self.model.get_secret(label=constants.AIRFLOW_RUNTIME_SECRET_LABEL)
```
shayancanonical: the two calls to `model.get_secret()` are redundant here

suggestion:
- create property for `peer_application_data`
- refactor `get_keys_secret()` to be idempotent to retrieve the secret. this method can be called in `config_generator.py` when rendering the `airflow_config.j2`
- make sure to call `_ensure_airflow_keys_generated()` in the reconcile method

```
constants.AIRFLOW_KEYS_SECRET = "airflow_keys_secret_id"
constants.AIRFLOW_KEYS_SECRET_LABEL = "airflow_keys_secret"

@property
def peer_application_data(self) -> ops.RelationData:
    return self._peer_relation.data[self.app]

def get_keys_secret(self) -> ops.Secret:
    keys_secret_id = self.peer_application_data.get(constants.AIRFLOW_KEYS_SECRET)
    if keys_secret_id:
        try:
            return self.model.get_secret(id=keys_secret_id, label=constants.AIRFLOW_KEYS_SECRET_LABEL)
        except (ops.SecretNotFoundError, ops.ModelError):
            logger.error("Issue retrieving secret hosting airflow keys")   
            raise ExitWithStatusError("Issue retrieving secret hosting airflow keys", ops.BlockedStatus)

    try:
        keys_secret = self.appl.add_secret(content, label=constants.AIRFLOW_KEYS_SECRET_LABEL)
        self.peer_application_data[constants.AIRFLOW_KEYS_SECRET] = keys_secret.id
    except ops.ValueError:
        logger.error("Issue adding secret for airflow keys")
        raise ExitWithStatusError("Issue adding secret for airflow keys", ops.BlockedStatus)

def _reconcile(self, event) -> None:
    ...
    self._ensure_airlfow_keys_generated()
    ...
```

jananisenthilkumar291: Agreed, addressed in commit [abefe8a](https://github.com/canonical/airflow-coordinator-k8s-operator/pull/31/commits/abefe8a51360ba02fb08b62f4797fb7cd9d7e49b)

src/charm.py
jananisenthilkumar291: Makes sense, addressed in [abefe8a](https://github.com/canonical/airflow-coordinator-k8s-operator/pull/31/commits/abefe8a51360ba02fb08b62f4797fb7cd9d7e49b)

src/charm.py
jananisenthilkumar291: Good point, Addressed in [abefe8a](https://github.com/canonical/airflow-coordinator-k8s-operator/pull/31/commits/abefe8a51360ba02fb08b62f4797fb7cd9d7e49b)

src/config_generator.py
```diff
@@ -56,6 +56,10 @@ def _sql_alchemy_connection_string(self) -> str:
     @property
     def sensitive_config_values(self) -> dict[str, str]:
         """All sensitive values that will be included in the Airflow config template."""
+        keys_content = self._charm.get_keys_secret().get_content()
         return {
             "sql_alchemy_connection_string": self._sql_alchemy_connection_string,
+            "secret_key": keys_content["secret-key"],
+            "jwt_secret": keys_content["jwt-secret"],
+            "fernet_key": keys_content["fernet-key"],
```
DnPlas: Please prepend these values with the section they correspond to, so they have the following format: section__key (exactly the same as stated in the [docs](https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html#secret-key)). For example secret_key will become "api__secret_key".

We need this because in future integrations with the k8s executor and others we need to transform these into env vars. cc @shayancanonical this is what we talked about a couple weeks ago.

jananisenthilkumar291: Good point!

Updated in [59c1e8a](https://github.com/canonical/airflow-coordinator-k8s-operator/pull/31/commits/59c1e8ae0c2f3ccbee8f4c87d0e5ac3d2e69deb5)[59c1e8a](https://github.com/canonical/airflow-coordinator-k8s-operator/pull/31/commits/59c1e8ae0c2f3ccbee8f4c87d0e5ac3d2e69deb5)

src/templates/airflow_config.j2
```diff
@@ -1,10 +1,15 @@
...
 load_examples = False
+fernet_key = {{ core__fernet_key }}
 
 [database]
 sql_alchemy_conn = {{ sql_alchemy_connection_string }}
 
 [api]
 base_url = {{ api_server_base_url }}
 port = {{ api_server_port }}
+secret_key = {{ api__secret_key }}
```
shayancanonical, line 10: nit: please update the remaining template variables in this file to follow the same format (we don't want 2 disjoint jinja variable naming conventions)

src/charm.py
```diff
@@ -128,16 +130,67 @@ def _peer_relation(self) -> ops.Relation:
...
+        except ValueError:
+            logger.error(constants.AIRFLOW_KEYS_SECRET_ADD_ERROR_MESSAGE)
+            raise ExceptionWithStatusError(
+                constants.AIRFLOW_KEYS_SECRET_ADD_ERROR_MESSAGE, ops.BlockedStatus
+            )
+
+    def get_keys_secret(self) -> ops.Secret:
+        """Retrieve the Juju application secret containing airflow keys."""
+        keys_secret_id = self.peer_application_data.get(constants.AIRFLOW_KEYS_SECRET)
+        if keys_secret_id:
```
shayancanonical: nit, prefer the other way

```
if not keys_secret_id:
    raise ExitWithStatusError

<code that assumes valid keys_secret_id>
```

it is far easier to trace the code as the consequence of not having the `keys_secret_id` is right next to the check for if it exists (rather than after an `if` block)

shayancanonical (commented):
looks great! 2 more minor nits - ready to approve after these are addressed

src/templates/airflow_config.j2
jananisenthilkumar291: Done. [2c782e0](https://github.com/canonical/airflow-coordinator-k8s-operator/pull/31/commits/2c782e085b4c8df3231fd34f5e720f105fbd7bb5)

src/charm.py
jananisenthilkumar291: Good point! Here's an update [2c782e0](https://github.com/canonical/airflow-coordinator-k8s-operator/pull/31/commits/2c782e085b4c8df3231fd34f5e720f105fbd7bb5)

shayancanonical (dismissed):
awesome work! there are now merge conflicts, please resolve and ping for re-approval after

shayancanonical (approved):

PR #32, opened 2026-03-20
[WF-335] Add relation with s3 interface to allow remote s3 dag bundles

shayancanonical:
## Description

- Add requires relation with `s3` interface
- Import s3 integrator charm lib
- For each new observed s3 connection information (from relation) run `airflow connections add` to create connection. For each removed relation, run `airflow connections delete`. this is done in `_reconcile_dag_bundle_remote_connections`. use peer relation to map relation id to the airflow connection for this relation
- Update config generator to include a s3 dag bundle entry for each valid s3 relation connection info
-  refactor command executor to instead have methods that return relevant `ops.pebble.ExecProcess` to execute. the process orchestration and error handling is done in a custom decorator that is used by all such methods that need to run `pebble exec`

---
## Testing instructions

please refer to justfiles in to set up microceph and deploy charms locally to test s3 dag bundles end-to-end: https://github.com/shayancanonical/tools 

install and configure S3 bucket locally with microceph:
```
1. just microceph-setup (this will install and setup microceph locally)

this will create a microceph user `test` (with necessary certs in `~/microceph_certs/` directory with relevant certs/keys to connect to microceph)

the s3 endpoint will be `https://$(just microceph-node-ip)`

sudo snap install aws-cli --classic

just s3-create-bucket mybucket foo bar (will create bucket 'mybucket' with access-key 'foo' and secret-key 'bar')
just s3-create-bucket anotherbucket foo2 bar2 (will create bucket 'anotherbucket' with access-key 'foo2' and secret-key 'bar2')

just s3-copy-into-bucket ./dag.py mybucket/dags/dag.py foo bar
just s3-copy-into-bucket ./dag2.py anotherbucket/dags/dag.py foo2 bar2
```
--

Deploy charms and test S3 DAG bundles

```
Ensure your airflow rocks repo locally has has the following PR changes until merge: https://github.com/canonical/airflow-rocks/pull/22
Ensure your airflow core repo locally has the following PR changes until merge: https://github.com/canonical/airflow-core-operators/pull/35
^ These are analogous changes required to test s3 dag bundles end-to-end locally

just airflow-deploy --model airflow --airflow-rock --coordinator-charm --api-server-charm --scheduler-charm --dag-processor-charm --triggerer-charm (adding the options indicates build of rock/charm from local repo, staged in ~/rocks and ~/charms. ensure proper repo directories in justfile)

juju ssh --container airflow-scheduler airflow-scheduler-k8s/0 cat /opt/airflow/airflow.cfg
^ ensure no `dag_processor.dag_bundle_config_list` config

just airflow-add-s3-bundle --bucket=mybucket --access-key=foo --secret-key=bar --path=dags/ --endpoint="https://$(just microceph-node-ip)" --tls-ca-chain-filepath=~/microceph_certs/ca.crt --app-name=s3-dags1

just airflow-add-s3-bundle --bucket=anotherbucket --access-key=foo2 --secret-key=bar2 --path=dags/ --endpoint="https://$(just microceph-node-ip)" --tls-ca-chain-filepath=~/microceph_certs/ca.crt --app-name=s3-dags2

juju ssh --container airflow-scheduler airflow-scheduler-k8s/0 cat /opt/airflow/airflow.cfg
^ ensure valid `dag_processor.dag_bundle_config_list` config
```

- I will actively keep cleaning just targets (especially microceph related) as the PR is being reviewed
---
## Checklist (all that apply)
- [ ] Documentation updated


src/charm.py
```diff
@@ -311,6 +364,72 @@ def _perform_checks(self) -> None:
...
+
+            if errorneous_relation_ids:
+                raise ExceptionWithStatusError(
+                    constants.INVALID_S3_RELATIONS_MESSAGE_TEMPLATE.format(
+                        relation_ids=", ".join(errorneous_relation_ids)
+                    ),
+                    ops.BlockedStatus,
+                )
+
+    def _reconcile_dag_bundle_remote_connections(self) -> None:
```
DnPlas: (nit, non-blocking) It seems to me like this method is doing three different operations:

1. Adding connections for the new s3 integrators that provide their connection info
2. Removing connections for s3 integrators that are not integrated any more
3. Update the statefulState _airflow_connection_ids with a "new_value"

Readability can be definitely improved if we abstract at least 1 and 2 to a helper method. We also benefit when unit testing these operations. I imagine something like:

```python3
  def _reconcile_dag_bundle_remote_connections(self) -> None:                                           
      if not self.unit.is_leader():
          return

      airflow_connections_to_update = {}
      self._create_or_update_s3_connections(airflow_connections_to_update)
      self._delete_stale_connections(airflow_connections_to_update)                                                           
   
      for connection_id, new_value in airflow_connections_to_update.items():                                                      
          self._airflow_connection_ids[connection_id] = new_value
```

src/charm.py
```diff
@@ -311,6 +364,72 @@ def _perform_checks(self) -> None:
...
+                    tls_ca_chain=connection_info.get("tls-ca-chain", []),
+                )
+
+                airflow_connections_to_update[connection_id] = connection_id
+
+        for airflow_connection_id in self._airflow_connection_ids:
+            connection_id_pattern = r"([\w_]+)_relation_(\d+)_connection"
+            match = re.match(connection_id_pattern, airflow_connection_id)
+
+            relation_id = int(match.group(2))
```
DnPlas: This will raise if there is no match, let's add a guardrail for that case.

src/charm.py
```diff
@@ -201,6 +208,52 @@ def _db_migration_ran(self, value: bool) -> None:
...
+                super().__init__(*args, **kwargs)
+
+                for key, value in json.loads(
+                    self.peer_application_data.get("airflow_connection_ids", "{}")
+                ).items():
+                    cls.data[key] = value
+
+            def __setitem__(cls, key, value):  # noqa: N805
+                if not value:
+                    cls.data.pop(key)
```
DnPlas: I believe this will raise if the key is not in cls.data.

src/charm.py
```diff
@@ -311,6 +364,72 @@ def _perform_checks(self) -> None:
...
+                self._command_executor.add_airflow_s3_connection(
+                    connection_id,
+                    connection_info["access-key"],
+                    connection_info["secret-key"],
+                    region=connection_info.get("region"),
+                    endpoint=connection_info.get("endpoint"),
+                    tls_ca_chain=connection_info.get("tls-ca-chain", []),
+                )
+
+                airflow_connections_to_update[connection_id] = connection_id
```
DnPlas: Should this be connection_info instead?

What's happening right now is that we are updating this temporal dictionary to {connection_id: connection_id}, which will cause later operations to not behave as we expect. In L431 when we do:

```python3
    for airflow_connection_id, new_value in airflow_connections_to_update.items():
            self._airflow_connection_ids[airflow_connection_id] = new_value
```

We are not actually giving it the connection_info we expect, as `new_value` is either:                                                                                  
  - "s3_relation_5_connection"
  - None

src/command_executor.py
```diff
@@ -32,57 +34,147 @@ class CommandExecutionResult(typing.NamedTuple):
...
+                "--conn-password",
+                secret_access_key,
+                *extras_options,
+            ],
+            user=constants.WORKLOAD_USER,
+            group=constants.WORKLOAD_GROUP,
+        )
+
+    @execute_pebble_exec_process
+    def delete_airflow_connection(self, connection_id: str) -> CommandExecutionResult:
```
DnPlas: Acording to the docs, I think this should be:

```suggestion
    def delete_airflow_connection(self, connection_id: str) -> ops.pebble.ExecProcess:
```

src/command_executor.py
```diff
@@ -32,57 +34,147 @@ class CommandExecutionResult(typing.NamedTuple):
...
+
+        if tls_ca_chain:
+            tls_ca_path = f"/opt/airflow/connection_certs/{connection_id}.pem"
+            try:
+                self._container.push(
+                    path=tls_ca_path,
+                    source="\n".join(tls_ca_chain),
+                    make_dirs=True,
+                    user=constants.WORKLOAD_USER,
+                    group=constants.WORKLOAD_USER,
```
DnPlas:
```suggestion
                    group=constants.WORKLOAD_GROUP,
```

The reason this is not failing is because they are both the same, but that's not always the case.

src/charm.py
```diff
@@ -201,6 +208,52 @@ def _db_migration_ran(self, value: bool) -> None:
         # We need to cast the bool to str, Juju relation data bags can only store strings
         self.peer_application_data["db_migration_ran"] = str(value).lower()
 
+    @functools.cached_property
+    def _airflow_connection_ids(self) -> dict[str, str]:
```
DnPlas: (non-blocking, but keep it mind) there is a risk that `peer_application_data` (a property that we added a couple months ago) changes the same databag as `_airflow_connection_ids`.

Nothing prevents other code from writing to that key directly and desyncing the two. Consider either making peer_application_data private, adding a comment marking exclusive ownership, or maybe something else.

src/charm.py
```diff
@@ -311,6 +364,72 @@ def _perform_checks(self) -> None:
...
+                if relation.id not in self.s3_connections
+            ]
+
+            if errorneous_relation_ids:
+                raise ExceptionWithStatusError(
+                    constants.INVALID_S3_RELATIONS_MESSAGE_TEMPLATE.format(
+                        relation_ids=", ".join(errorneous_relation_ids)
+                    ),
+                    ops.BlockedStatus,
+                )
```
DnPlas: Would it make sense to extract this into a self contained helper method? That way we can get rid of the C901 exception from above.

src/charm.py
```diff
@@ -311,6 +364,72 @@ def _perform_checks(self) -> None:
...
+
+                airflow_connections_to_update[connection_id] = connection_id
+
+        for airflow_connection_id in self._airflow_connection_ids:
+            connection_id_pattern = r"([\w_]+)_relation_(\d+)_connection"
+            match = re.match(connection_id_pattern, airflow_connection_id)
+
+            relation_id = int(match.group(2))
+
+            # TODO also ensure relation_id false-y in git connections
```
DnPlas: Let's add an issue and add it here as a comment

DnPlas (commented):
Thanks @shayancanonical , I wasn't able to follow the guide and therefore to manually test it. We're missing a target in the [justfile](https://github.com/shayancanonical/tools/blob/main/justfile):

```
just set-up-microceph
just install-and-bootstrap-microceph
error: no matching snaps installed
microceph (squid/stable) 19.2.3+snapafb6a3a0cb from Canonical✓ installed

+-----------+---------+
|   PATH    | STATUS  |
+-----------+---------+
| loop,1G,3 | Success |
+-----------+---------+
just setup-microceph-certs
error: Justfile does not contain recipe `setup-microceph-certs`
error: Recipe `set-up-microceph` failed on line 94 with exit code 1
```

src/charm.py
shayancanonical: addressed + refactored in a2a2c8ee142ffb966605ed88e1bcc3d99024ef16

src/charm.py
shayancanonical: removed regexp matching in a2a2c8ee142ffb966605ed88e1bcc3d99024ef16. had a brain freeze and was mapping connection_id to connection_id instead of relation_id to connection_id

src/charm.py
shayancanonical: added `cls.data.pop(key, None)` to make it safe in a2a2c8ee142ffb966605ed88e1bcc3d99024ef16

src/charm.py
shayancanonical: doh, meant to map `relation_id` to `connection_id`. had a brain freeze when working on this and did something silly and round-about
addressed in a2a2c8ee142ffb966605ed88e1bcc3d99024ef16

src/command_executor.py
shayancanonical: addressed in a2a2c8ee142ffb966605ed88e1bcc3d99024ef16

src/command_executor.py
shayancanonical: addressed in a2a2c8ee142ffb966605ed88e1bcc3d99024ef16

src/charm.py
shayancanonical: made the property (`peer_application_data`) private in a2a2c8ee142ffb966605ed88e1bcc3d99024ef16

within the peer data bag, `_airflow_connection_ids` updates/writes a key that i would hope no method in this class now mistakenly overwrites

src/charm.py
shayancanonical: extracted in a2a2c8ee142ffb966605ed88e1bcc3d99024ef16

src/charm.py
shayancanonical: removed the comment in a2a2c8ee142ffb966605ed88e1bcc3d99024ef16. we already have a jira ticket (in queue to work on next)

shayancanonical:
@DnPlas thanks for the feedback. the justfile targets were hastily put together while experimenting and i did not have the chance to refine and test further until this evening (though i had hoped that the target contents would temporarily give a primer on how to set up microceph and achieve the objective of getting a review quicker)

refined and tested the microceph related targets in https://github.com/shayancanonical/tools
(`just setup-microceph` will now do the trick. also added `just create-bucket`, `just list-bucket` and `just copy-into-bucket` to test out `aws s3` commands with the locally running microceph. there'll be further polish of the targets as i continually keep using them, please feel free to share suggestions to improve if you experience pain points)

tests/unit/test_charm.py
```diff
@@ -366,6 +371,146 @@ def test_db_migration_failure(context, state, mock_command_executor, workload_co
...
+                    }
+                    for relation_id, connection_info in {
+                        s3_integrator_relation.id: s3_integrator_relation.remote_app_data,
+                        s3_integrator_relation2.id: s3_integrator_relation2.remote_app_data,
+                    }.items()
+                ]
+            )
+
+    def test_valid_s3_relations_non_leader(self, context, state):
+        """Non-leader coordiantor units no-op."""
```
jananisenthilkumar291: nit
```suggestion
        """Non-leader coordinator units no-op."""
```

src/charm.py
```diff
@@ -311,6 +366,97 @@ def _perform_checks(self) -> None:
...
+
+        s3_relation_connection_ids = [
+            f"s3_relation_{relation_id}_connection" for relation_id in self.s3_connections
+        ]
+
+        for airflow_connection_id in [connection["conn_id"] for connection in airflow_connections]:
+            if airflow_connection_id not in s3_relation_connection_ids:
+                logger.info(f"Deleting Airflow connection {airflow_connection_id}")
+
+                self._command_executor.delete(airflow_connection_id)
```
jananisenthilkumar291:
```suggestion
                self._command_executor.delete_airflow_connection(airflow_connection_id)
```

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
```diff
@@ -269,12 +269,15 @@ class AirflowCoordinatorProviderModel(data_interfaces.BaseCommonModel):
     # (str from coordinator) and structured config dicts (from executor, deserialized
     # by data_interfaces' get_data). This will change when we refactor the library
     # to a much more generic one.
-    config_template: str | dict | None = pydantic.Field(default=None)
-    kubernetes_executor_pod_spec: str | None = pydantic.Field(default=None)
-    sensitive_data: SensitiveDataSecretStr = pydantic.Field(default=None)
-    secret_sensitive_data: data_interfaces.SecretString | None = pydantic.Field(default=None)
+    config_template: str | dict | None = None
```
DnPlas, line 272: yeh, this change is making it more evident that we need this library to be absolutely generic, the more "configurators" and/or integrators we add.

@shayancanonical please file an issue where we track this enhancement.

src/charm.py
DnPlas: Let's add a unit test for this as it was not caught.

```suggestion
                self._command_executor.delete_airflow_connection(airflow_connection_id)
```

src/charm.py
```diff
@@ -311,6 +366,97 @@ def _perform_checks(self) -> None:
...
+        airflow_connections = json.loads(
+            self._command_executor.list_airflow_connections().stdout or "{}"
+        )
+
+        s3_relation_connection_ids = [
+            f"s3_relation_{relation_id}_connection" for relation_id in self.s3_connections
+        ]
+
+        for airflow_connection_id in [connection["conn_id"] for connection in airflow_connections]:
+            if airflow_connection_id not in s3_relation_connection_ids:
```
DnPlas, line 506: Is it possible that this will eliminate connections that we actually want to keep (e.g. git connections or built-in ones)?

src/charm.py
```diff
@@ -311,6 +366,97 @@ def _perform_checks(self) -> None:
...
+                    f"Connection info for {connection_id} changed. Updating Airflow connection"
+                )
+
+                self._command_executor.add_airflow_s3_connection(
+                    connection_id,
+                    connection_info["access-key"],
+                    connection_info["secret-key"],
+                    region=connection_info.get("region"),
+                    endpoint=connection_info.get("endpoint"),
+                    tls_ca_chain=connection_info.get("tls-ca-chain", []),
```
DnPlas: What's the actual key `tls-ca-chain` or `tls_ca_chain`? There is some inconsistency throughout the file.

src/charm.py
```diff
@@ -193,13 +197,47 @@ def get_keys_secret(self) -> ops.Secret:
...
+            if relation
+            and (
+                (connection_info := self._s3_requires.get_storage_connection_info(relation))
+                or connection_info == {}
+            )
+            and (
+                not connection_info
+                or all(key in connection_info for key in ["bucket", "access-key", "secret-key"])
+            )
+        }
```
DnPlas, line 261: (nit) While I believe this is correct, this statement is really hard to read and understand. I would consider a helper method, perhaps or split some of these operations inside the property.

src/charm.py
```diff
@@ -158,15 +162,15 @@ def _ensure_airflow_keys_generated(self) -> None:
...
         """
-        if self.peer_application_data.get(constants.AIRFLOW_KEYS_SECRET):
+        if self._peer_application_data.get(constants.AIRFLOW_KEYS_SECRET):
             return
 
         try:
             content = self._generate_keys_secret_content()
             keys_secret = self.app.add_secret(content, label=constants.AIRFLOW_KEYS_SECRET_LABEL)
             if not keys_secret.id:
                 raise ValueError("Secret created without an ID")
```
DnPlas, line 206: (non-blocking) not related to this PR, but it caught my eye - why are we raising a valueError here? I thought the pattern was to raise the ExceptionWithStatusError.

src/charm.py
```diff
@@ -311,6 +366,97 @@ def _perform_checks(self) -> None:
                 constants.MISMATCHED_WORKLOAD_IMAGE_HASHES_MESSAGE, ops.BlockedStatus
             )
 
+        self._perform_potential_s3_connection_checks()
+
+    def _create_or_update_s3_connections(self) -> None:
+        """Create or update s3 connections that have changed."""
+        airflow_connections = json.loads(
+            self._command_executor.list_airflow_connections().stdout or "{}"
```
DnPlas: (nit) I don't think that you can default to {} when the return type of airflow connections list is a JSON array. I looked at the src code of the CLI to confirm this ([see here](https://github.com/apache/airflow/blob/main/airflow-ctl/src/airflowctl/ctl/console_formatting.py#L111)). The output of the command will be something like:

```
[
  {"key1": "value1"},
  {"key2": "value2"}
]
```

so...it seems to me like the reason this is working now is because the dict is empty, and in later processing we are not iterating over any key, but, if you want to be more precise, it'll be a [].

tests/unit/test_command_executor.py
DnPlas, line 1: I think there are a couple methods that are not tested, consider adding them for better coverage and catching errors like the method name delete():

* _create_or_update_s3_connections_create_or_update_s3_connections
* _delete_stale_connections (as mentioned above)
* _has_s3_connection_changed
* _perform_potential_s3_connection_checks
* _reconcile_dag_bundle_remote_connections

We can have (at least) some assertions on the methods that are supposed to run.

src/command_executor.py
```diff
@@ -32,57 +34,157 @@ class CommandExecutionResult(typing.NamedTuple):
...
+            group=constants.WORKLOAD_GROUP,
+        )
+
+    @execute_pebble_exec_process
+    def list_airflow_connections(self) -> ops.pebble.ExecProcess:
+        """List all Airflow connections."""
+        return self._container.exec(
+            ["airflow", "connections", "list", "--output", "json"],
+            user=constants.WORKLOAD_USER,
+            group=constants.WORKLOAD_GROUP,
```
DnPlas, line 188: I see this repeated across the charm code:

```
        airflow_connections = json.loads(
            self._command_executor.list_airflow_connections().stdout or "{}"
        )
```

would it make sense to do that here instead ? so, instead of returning the execProcess you'd return the dict, perhaps that way it'd be clearer what the type of this method is and it'd be easier to run operations over it (such as what you are doing in _create_or_update_s3_connections or _has_s3_connection_changed.

src/command_executor.py
```diff
@@ -32,57 +34,157 @@ class CommandExecutionResult(typing.NamedTuple):
...
+        """
+        self._container = container
+
+    @execute_pebble_exec_process
+    def run_db_migrate(self) -> ops.pebble.ExecProcess:
+        """Execute the 'airflow db migrate' command."""
+        return self._container.exec(
+            ["airflow", "db", "migrate"],
+            environment={
+                "AIRFLOW_HOME": "/opt/airflow",
```
DnPlas: (nit) Should we use the value from constants instead?

src/charm.py
```diff
@@ -311,6 +366,97 @@ def _perform_checks(self) -> None:
...
+                    connection_info["secret-key"],
+                    region=connection_info.get("region"),
+                    endpoint=connection_info.get("endpoint"),
+                    tls_ca_chain=connection_info.get("tls-ca-chain", []),
+                )
+
+    def _delete_stale_connections(self) -> None:
+        """Delete stale s3 connections (s3 connections without relations)."""
+        airflow_connections = json.loads(
+            self._command_executor.list_airflow_connections().stdout or "{}"
```
DnPlas: Another guard we can add here is the .success check (similar to db migrate), that way we make sure the exceptions are propagated correctly and we set the unit to some status if there are failures.

src/charm.py
```diff
@@ -355,6 +501,7 @@ def _reconcile(self, event: ops.EventBase) -> None:
         try:
             self._perform_checks()
             self._ensure_airflow_keys_generated()
+            self._reconcile_dag_bundle_remote_connections()
```
DnPlas: To add/list/delete connections, don't we need the airflow config file which is only generated after L506? If that's the case, maybe we need to move this operation beneath the wirte_airflow_config() call.

DnPlas: Also, if we do move the call after L506, we may not even need to check for leadership, and even db migration.

src/command_executor.py
```diff
@@ -32,57 +34,157 @@ class CommandExecutionResult(typing.NamedTuple):
...
+                self._container.push(
+                    path=tls_ca_path,
+                    source="\n".join(tls_ca_chain),
+                    make_dirs=True,
+                    user=constants.WORKLOAD_USER,
+                    group=constants.WORKLOAD_GROUP,
+                )
+            except ops.pebble.PathError as e:
+                logger.error(f"Unexpected error pushing TLS CA chain: {e}")
+                raise CommandExecutionError(f"Failed to push TLS CA chain: {e}") from e
```
DnPlas: I'd probably push this out of this method, it seems self-contained enough and is not that related to adding the connection, but rather it's doing some processing of a file.

src/command_executor.py
```diff
@@ -32,57 +34,157 @@ class CommandExecutionResult(typing.NamedTuple):
...
+
+            extras["verify"] = tls_ca_path
+
+        extras_options = ["--conn-extra", json.dumps(extras)] if extras else []
+
+        # airflow connections add also updates existing connections
+        return self._container.exec(
```
DnPlas, line 149: Is AIRFLOW_HOME needed here?

DnPlas (commented):
Thanks @shayancanonical, left some comments.

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
shayancanonical: created https://github.com/canonical/airflow-coordinator-k8s-operator/issues/38

src/charm.py
shayancanonical: yep, the raised `ValueError` is caught right after (this code is in a try/except  block), and the except block will raise the `ExceptionWithStatusError`. the reason why a ValueError is raised here is because it fits + `self.app.add_secret()` can raise a value error too

tests/unit/test_charm.py
shayancanonical: addressed in 35d178406519c9a6f3366fc371f367535fece82f

src/charm.py
shayancanonical: addressed in 35d178406519c9a6f3366fc371f367535fece82f

shayancanonical: added unit test in 35d178406519c9a6f3366fc371f367535fece82f

src/charm.py
shayancanonical: oof, this would have been a disaster if not caught. addressed in 35d178406519c9a6f3366fc371f367535fece82f

src/charm.py
shayancanonical: use a dataclass to represent the S3 connection info (coming from upstream related s3-integrator app) for simplicity: 35d178406519c9a6f3366fc371f367535fece82f

src/charm.py
shayancanonical: simplified with use of dataclass in 35d178406519c9a6f3366fc371f367535fece82f

src/charm.py
shayancanonical: addressed by adding `parsed_output` to `CommandExecutionResult` if `stdout` is JSON parseable. should simplify by eliminating defaults in callers: 35d178406519c9a6f3366fc371f367535fece82f

tests/unit/test_command_executor.py
shayancanonical: added 2 unit tests (to ensure changed s3 relations get updated, and stale s3 connections due to removed relations get cleaned) in 35d178406519c9a6f3366fc371f367535fece82f

src/command_executor.py
shayancanonical: doing this by adding `parsed_output` to `CommandExecutionResult` so it is done in a unified way (inside the decorator): 35d178406519c9a6f3366fc371f367535fece82f

src/command_executor.py
shayancanonical: use from constants in 35d178406519c9a6f3366fc371f367535fece82f

src/charm.py
shayancanonical: i believe with accessing `parsed_output` (35d178406519c9a6f3366fc371f367535fece82f), there's no need for `success` check

if `not success`, a `CommandExecutionError` will get raised, which in turn will get caught in `_reconcile_dag_bundle_remote_connections` (which then raises `ExceptionWithStatusError`

src/charm.py
shayancanonical: moved in 35d178406519c9a6f3366fc371f367535fece82f. previously, `_reconcile_dag_bundle_remote_connections` had a guard against `_db_migration_ran == false`. though, the proposal in this comment is cleaner, so switched to proposed approach

src/command_executor.py
shayancanonical: pushed out into the caller (in `_create_or_update_s3_connections()`: 35d178406519c9a6f3366fc371f367535fece82f

src/command_executor.py
shayancanonical: added `AIRFLOW_HOME` to all `exec()` calls in this class: 35d178406519c9a6f3366fc371f367535fece82f

src/charm.py
```diff
@@ -311,6 +399,129 @@ def _perform_checks(self) -> None:
...
+
+            if not filtered_airflow_connections:
+                return True
+
+            airflow_connection = filtered_airflow_connections[0]
+
+            try:
+                tls_ca_chain = (
+                    self._container.pull(
+                        f"/opt/airflow/connection_certs/{connection_id}.pem",
```
DnPlas: I think you want to make this `f"{AIRFLOW_HOME/connection_certs<etc.>"`

DnPlas (commented):
One teeny tiny nit, other than that LGTM

src/charm.py
shayancanonical: addressed in c496024afc249a240985c573cfbb72531044490a

DnPlas (dismissed):
Thanks @shayancanonical, I wasn't able to test it, the just targets are off, but I saw it on the demo, gj!

DnPlas (approved):

PR #37, opened 2026-03-31
[WF-316] Support for git dag bundles

shayancanonical:
## Description

- Adds relation endpoint (`git`) to support upstream git-integrator relation
- Validates errorneous git-integrator data
- For all git connections with authentication methods, adds Airflow connection. Also cleans up stale git connections in the database (for relations removed)
- Includes git connections as dag bundles
- Implements scenario tests
- Imports git-integrator charm lib

---
## Testing instructions

Deploy charms and test git DAG bundles

```
Ensure your airflow rocks repo locally has has the following PR changes until merge: https://github.com/canonical/airflow-rocks/pull/22
Ensure your airflow core repo locally has the following PR changes until merge: https://github.com/canonical/airflow-core-operators/pull/35
^ These are analogous changes required to test s3 dag bundles end-to-end locally

just airflow-deploy airflow-model --airflow-rock --coordinator-charm --api-server-charm --scheduler-charm --dag-processor-charm --triggerer-charm (adding the options indicates build of rock/charm from local repo, staged in ~/rocks and ~/charms. ensure proper repo directories in justfile)

juju ssh --container airflow-scheduler airflow-scheduler-k8s/0 cat /opt/airflow/airflow.cfg
^ ensure no `dag_processor.dag_bundle_config_list` config

just airflow-add-git-bundle "https://github.com/apache/airflow" --path="airflow-core/src/airflow/example_dags" --tracking-ref="main" --app-name git-dag1

juju ssh --container airflow-scheduler airflow-scheduler-k8s/0 cat /opt/airflow/airflow.cfg
^ ensure valid `dag_processor.dag_bundle_config_list` config

the dags from airflow's repo will not be loaded into the system (verifiable from the UI by triggering a DAG)
```

This PR must be merged after the merge of #32 (after which the base branch will be updated to `track/3.1`)

---
## Checklist (all that apply)
- [ ] Documentation updated


src/charm.py
```diff
@@ -304,6 +307,24 @@ def _perform_potential_s3_connection_checks(self) -> None:
...
+    def _perform_potential_git_connection_checks(self) -> None:
+        """Checks validity of all present git connections."""
+        if self._git_requires.relations:
+            errorneous_relation_ids = [
+                str(relation.id)
+                for relation in self._git_requires.relations
+                if relation.id not in self._git_requires.get_git_connection_information()
+                and relation.data[relation.app]
+            ]
+
```
DnPlas, line 372: (nit, non-blocking) seems like a candidate for a health check that the interface can do

src/charm.py
```diff
@@ -427,21 +449,124 @@ def _has_s3_connection_changed(
...
             self._command_executor.list_airflow_connections().stdout or "{}"
         )
 
-        s3_relation_connection_ids = [
+        airflow_connection_ids = [connection["conn_id"] for connection in airflow_connections]
+
+        def _has_git_connection_changed(
+            connection_id: str, git_provider_model: git.GitProviderModel
+        ) -> bool:
+            """Helper to determine if an git connection's info has changed."""
```
DnPlas: (nit, spelling mistake) 

```suggestion
            """Helper to determine if a git connection's info has changed."""
```

DnPlas: (nit) 

```suggestion
            """Return True if X, Y otherwise"""
```

src/charm.py
```diff
@@ -427,21 +449,124 @@ def _has_s3_connection_changed(
...
+
+            elif git_provider_model.authentication_method == git.AuthenticationMethodEnum.SSH:
+                # Existing airflow connection has credentials
+                if airflow_connection.get("login") or airflow_connection.get("password"):
+                    has_authentication_changed = True
+
+                if (
+                    airflow_connection["extra_dejson"].get("ssh_key")
+                    != git_provider_model.ssh_private_key
+                    or airflow_connection.get["extra_dejson"].get("strict_host_key_checking")
```
DnPlas: typo?

```suggestion
                    or airflow_connection["extra_dejson"].get("strict_host_key_checking")
```

src/charm.py
```diff
@@ -427,21 +449,124 @@ def _has_s3_connection_changed(
...
+            )
+
+        for (
+            relation_id,
+            git_provider_model,
+        ) in self._git_requires.get_git_connection_information().items():
+            if git_provider_model.authentication_method is None:
+                logger.debug(
+                    f"Skipping Airflow connection creation for relation {relation_id}"
+                    " since it has not authentication"
```
DnPlas: (nit)

```suggestion
                    " since it has no authentication"
```

src/charm.py
```diff
@@ -427,21 +449,124 @@ def _has_s3_connection_changed(
...
+            )
+
+            if not filtered_airflow_connections:
+                return True
+
+            airflow_connection = filtered_airflow_connections[0]
+
+            has_authentication_changed = False
+
+            if (
```
DnPlas: Consider using ORs instead of multiple if statements (perhaps this is more a stylistic choice). I imagine something like:

```
  if method == CREDENTIALS:
      has_authentication_changed = (
          airflow_connection["extra_dejson"].get("ssh_key")
          or airflow_connection["extra_dejson"].get("strict_host_key_checking")
          or airflow_connection.get("login") != git_provider_model.credentials_username
          or airflow_connection.get("password") != git_provider_model.credentials_personal_access_token
      )
...
```

Please also consider adding an `else` clause at the end to avoid the scenario of method being neither CREDENTIALS or SSH

DnPlas (commented):
Thanks @shayancanonical, left some nits, nothing too concerning atm, I'll test after s3 changes are merged.

src/charm.py
shayancanonical: agreed, added issue to `git-integrator` repo to add this to the charm lib: https://github.com/canonical/git-integrator/issues/11

src/charm.py
shayancanonical: suggestion makes code much cleaner - made change in 14233529e98f8a90ab03c63bae2fe4626f11f4ce

src/charm.py
shayancanonical: addressed in 14233529e98f8a90ab03c63bae2fe4626f11f4ce

src/charm.py
shayancanonical: addressed in 14233529e98f8a90ab03c63bae2fe4626f11f4ce (caught while implementing further unit tests for change in git relation data and removal of git relation data)

src/charm.py
shayancanonical: updated in 14233529e98f8a90ab03c63bae2fe4626f11f4ce

shayancanonical: updated in 14233529e98f8a90ab03c63bae2fe4626f11f4ce

src/charm.py
```diff
@@ -514,8 +635,9 @@ def _reconcile_dag_bundle_remote_connections(self) -> None:
             return
 
         try:
-            self._create_or_update_s3_connections()
             self._delete_stale_connections()
+            self._create_or_update_s3_connections()
+            self._create_or_update_git_connections()
         except command_executor.CommandExecutionError:
             raise ExceptionWithStatusError(
                 "Internal issue of reconciling S3 airflow connections",
```
DnPlas: (nit)

```suggestion
                "Internal issue of reconciling S3/git airflow connections",
```

src/charm.py
```diff
@@ -484,6 +506,96 @@ def _has_s3_connection_changed(
...
+            filtered_airflow_connections = list(
+                filter(
+                    lambda connection: connection["conn_id"] == connection_id, airflow_connections
+                )
+            )
+
+            if not filtered_airflow_connections:
+                return True
+
+            airflow_connection = filtered_airflow_connections[0]
```
DnPlas: (nit) I think up to this point, this code is shared with the s3 check method, this could be abstracted to a single helper for both.

src/charm.py
```diff
@@ -484,6 +506,96 @@ def _has_s3_connection_changed(
...
 
+    def _create_or_update_git_connections(self) -> None:  # noqa: C901
+        """Create or update git connections that have changed."""
+        airflow_connections = self._command_executor.list_airflow_connections().parsed_stdout
+
+        if airflow_connections is None:
+            raise ExceptionWithStatusError(
+                constants.ISSUE_QUERYING_DATABASE_MESSAGE,
+                ops.BlockedStatus,
+            )
```
DnPlas: Similar case here than below, up to this point, this code is shared between s3 and git, consider abstracting to a helper method. It'd be good as a way to avoid listing the connections twice in a row (one for git, one for s3).

src/charm.py
```diff
@@ -484,6 +506,96 @@ def _has_s3_connection_changed(
...
+
+                logger.info(f"Adding new Airflow connection for {connection_id}")
+
+                self._command_executor.add_airflow_git_connection(
+                    connection_id,
+                    git_provider_model,
+                )
+
     def _delete_stale_connections(self) -> None:
         """Delete stale s3 connections (s3 connections without relations)."""
```
DnPlas: (nit) 
```suggestion
        """Delete stale s3/git connections (s3/git connections without relations)."""
```

or maybe make it more general?

src/charm.py
```diff
@@ -484,6 +506,96 @@ def _has_s3_connection_changed(
...
+
+        if airflow_connections is None:
+            raise ExceptionWithStatusError(
+                constants.ISSUE_QUERYING_DATABASE_MESSAGE,
+                ops.BlockedStatus,
+            )
+
+        airflow_connection_ids = [connection["conn_id"] for connection in airflow_connections]
+
+        def _has_git_connection_changed(
```
DnPlas: (nit, non-blocking) This is potentially an enhancement to the git library, it can be part of it to avoid making the charm code do this operation.

src/config_generator.py
```diff
@@ -130,12 +130,37 @@ def dag_processor_config(self) -> dict[str, dict[str, str]]:
             if connection_info
```
DnPlas: (nit, not blocking) tiny inconsistency on the naming, could it be s3_provider_model to match the git one? this can be done in a follow up PR, btw.

src/config_generator.py
```diff
@@ -130,12 +130,37 @@ def dag_processor_config(self) -> dict[str, dict[str, str]]:
...
+                    key: value
+                    for key, value in {
+                        "git_conn_id": f"git_relation_{relation_id}_connection"
+                        if git_provider_model.authentication_method is not None
+                        else None,
+                        "repo_url": git_provider_model.repository_url,
+                        "tracking_ref": git_provider_model.tracking_ref,
+                        "subdir": git_provider_model.path,
+                        "submodules": False,
+                        "prune_dotgit_folder": True,
```
DnPlas: Is this something we want to leave out of the user configurable options?

src/config_generator.py
```diff
@@ -130,12 +130,37 @@ def dag_processor_config(self) -> dict[str, dict[str, str]]:
...
+                    }.items()
+                    if value is not None
```
DnPlas: (nit, non-blocking) This could be something we can add to the s3 as well, consider abstracting some of the operations as much as possible to avoid repeating code and ensuring you have these sort of guardrails for both connections.

DnPlas (commented):
Thanks @shayancanonical, some nits here and there, but overall LGTM

src/charm.py
shayancanonical: updated in https://github.com/canonical/airflow-coordinator-k8s-operator/pull/37/commits/26bc8079bbf3d0c3ed1d6b9498669a79f3e5c594 + https://github.com/canonical/airflow-coordinator-k8s-operator/pull/37/commits/0b977d16e33899220358cc4b6fc174034c3943af

src/charm.py
shayancanonical: added helper methods to get all connections from db, connections of type (git/s3). will keep an eye and likely add a method for connection with a particular id. however, the main change was to use a `cached_property` to avoid re-executing `airflow connections list` (which then results in a round trip network request all the way to postgres). all connections are cached in `airflow_connections` property and we can keep filtering as we desire

https://github.com/canonical/airflow-coordinator-k8s-operator/pull/37/commits/26bc8079bbf3d0c3ed1d6b9498669a79f3e5c594 + https://github.com/canonical/airflow-coordinator-k8s-operator/pull/37/commits/0b977d16e33899220358cc4b6fc174034c3943af

src/charm.py
shayancanonical: updated in in https://github.com/canonical/airflow-coordinator-k8s-operator/pull/37/commits/26bc8079bbf3d0c3ed1d6b9498669a79f3e5c594 + https://github.com/canonical/airflow-coordinator-k8s-operator/pull/37/commits/0b977d16e33899220358cc4b6fc174034c3943af

src/charm.py
shayancanonical: moved connection related code to a separate file inin https://github.com/canonical/airflow-coordinator-k8s-operator/pull/37/commits/26bc8079bbf3d0c3ed1d6b9498669a79f3e5c594 + https://github.com/canonical/airflow-coordinator-k8s-operator/pull/37/commits/0b977d16e33899220358cc4b6fc174034c3943af

spoke with bikalpa in data platform tech talk - suggested that git lib host a dataclass representation of relation data. we talked our way into agreeing that change logic is a client concern, not a lib concern. with the linked commits, we now use a dataclass to make the change detection a bit cleaner

src/charm.py
shayancanonical: connections are cached in https://github.com/canonical/airflow-coordinator-k8s-operator/pull/37/commits/26bc8079bbf3d0c3ed1d6b9498669a79f3e5c594 + https://github.com/canonical/airflow-coordinator-k8s-operator/pull/37/commits/0b977d16e33899220358cc4b6fc174034c3943af

merged `_perform_potential_s3_connection_checks()` and `_perform_potential_git_connection_checks()` into a single method in 6a0c2e9afb5a706af63f38334f318d01f0698e6e

src/config_generator.py
shayancanonical: i changed to s3_connection_info to be more consistent with how the dataclass for the S3 relation data is named

6a0c2e9afb5a706af63f38334f318d01f0698e6e

src/config_generator.py
shayancanonical: cleaned up further in 6a0c2e9afb5a706af63f38334f318d01f0698e6e

i believe the changes make it easier to follow how the dag bundle configs are compiled from the respective s3 and git relation data

src/config_generator.py
shayancanonical: would you please be able to clarify how you mean? i added the two fixed values (`submodules=False` and `prune_dotgit_folder=True`) based on https://airflow.apache.org/docs/apache-airflow-providers-git/stable/bundles/index.html

they seemed sensible, as i think it might be good to avoid pulling down the `.git` directory and any submodules from a specified git connection. please lmk if you have any thoughts or if you think it might be easier to come back to this in the future and they should be left out for now

src/connection_manager.py
```diff
@@ -0,0 +1,303 @@
...
+            return False
+
+        if git_provider_model.authentication_method == git.AuthenticationMethodEnum.CREDENTIALS:
+            # Existing airflow connection has SSH parameters or
+            # credentials parameters changed
+            has_authentication_changed = (
+                airflow_connection.extra_dejson.get("private_key")
+                or airflow_connection.extra_dejson.get("strict_host_key_checking")
+                or airflow_connection.get.login != git_provider_model.credentials_username
+                or airflow_connection.get.password
```
DnPlas: since you introduced the connection data class, this is no longer valid

```suggestion
                or airflow_connection.login != git_provider_model.credentials_username
                or airflow_connection.password
```

Please make sure we add a unit test so this is not missed again

src/connection_manager.py
```diff
@@ -0,0 +1,303 @@
...
+
+    def get_airflow_connections_by_type(self, type: str) -> list[dict]:
+        """Airflow connections of a specific type from the database.
+
+        Examples of specified type can be "aws" and "git".
+        """
+        return list(
+            filter(lambda connection: connection["conn_type"] == type, self.airflow_connections)
+        )
+
```
DnPlas: Similar case here as above, should it be connection.conn_type?

src/connection_manager.py
```diff
@@ -0,0 +1,303 @@
...
+            connection_id = f"git_relation_{relation_id}_connection"
+
+            if self.has_connection_for_git_changed(connection_id, git_provider_model):
+                logger.info(
+                    f"Connection info for {connection_id} changed. Deleting old Airflow connection"
+                )
+
+                if connection_id in airflow_connection_ids:
+                    self._command_executor.delete_airflow_connection(connection_id)
```
DnPlas:
```suggestion
                    self._charm._command_executor.delete_airflow_connection(connection_id)
```

src/connection_manager.py
```diff
@@ -0,0 +1,303 @@
...
+
+class AirflowConnectionManager:
+    """Encapsulate business logic around management of Airflow connections."""
+
+    def __init__(self, charm: ops.CharmBase, container: ops.Container):
+        self._charm = charm
+        self._container = container
+
+    @functools.cached_property
+    def airflow_connections(self) -> list[AirflowConnection]:
```
DnPlas, line 92: I don't think this is exactly accurate as `AirflowConnection.from_airflow_connections_list_output()` may return None or an empty list. You can either change this method's return type or the `from_airflow_connections_list_output()` for better handling the returns. See my comment above.

src/connection_manager.py
```diff
@@ -0,0 +1,303 @@
...
+    extra_dejson: dict = dataclasses.field(default_factory=dict)
+
+    @classmethod
+    def from_airflow_connections_list_output(cls, data: list[dict]):
+        """Instantiate from parsed json output of `airflow connections list`."""
+        if not data:
+            return []
+
+        if not all(connection.get("conn_type") for connection in data):
+            return None
```
DnPlas: Should this be treated as an error, rather? Perhaps not having a "conn_type" is something that should be raised as an exception.

src/connection_manager.py
```diff
@@ -0,0 +1,303 @@
...
+
+        Must be run to ensure re-fetch/re-compute of cached data.
+        """
+        del self.__dict__["airflow_connections"]
+
+    def get_airflow_connections_by_type(self, type: str) -> list[dict]:
+        """Airflow connections of a specific type from the database.
+
+        Examples of specified type can be "aws" and "git".
+        """
```
DnPlas: nit:

```suggestion
        """Airflow connections of a specific type from the database.

        Args:
            type (str): the type of connection to get, accepted values are: x, y, z.
        """
```

src/connection_manager.py
```diff
@@ -0,0 +1,303 @@
...
+        return AirflowConnection.from_airflow_connections_list_output(raw_connections)
+
+    def refresh(self):
+        """Refresh cached properties of this class.
+
+        Must be run to ensure re-fetch/re-compute of cached data.
+        """
+        del self.__dict__["airflow_connections"]
+
+    def get_airflow_connections_by_type(self, type: str) -> list[dict]:
```
DnPlas: Are we at all using this method?

src/connection_manager.py
```diff
@@ -0,0 +1,303 @@
...
+        Examples of specified type can be "aws" and "git".
+        """
+        return list(
+            filter(lambda connection: connection["conn_type"] == type, self.airflow_connections)
+        )
+
+    def has_connection_for_s3_changed(
+        self, connection_id: str, connection_info: S3ConnectionInfo
+    ) -> bool:
+        """Validator for change of Airflow AWS connection compared to S3 relation data."""
```
DnPlas: nit:

```suggestion
        """Returns True if x; otherwise False
        
        Args:
            connection_id (str):
            connection_info (S3ConnectionInfo):
        """
```

src/connection_manager.py
```diff
@@ -0,0 +1,303 @@
...
+
+        s3_relation_connection_ids = [
+            f"s3_relation_{relation_id}_connection"
+            for relation_id in self._charm.s3_relation_connections
+        ]
+
+        git_relation_connection_ids = [
+            f"git_relation_{relation_id}_connection"
+            for relation_id in self._charm._git_requires.get_git_connection_information()
+        ]
```
DnPlas, line 276: question: the naming is some sort of guard against the charm removing connections that were created by users via the web UI, is that correct?

DnPlas (commented):
Thanks @shayancanonical, some comments.

src/connection_manager.py
shayancanonical: fixed in 105603853b0d9b3ba3f421ad9b34d4851eeef15f

src/connection_manager.py
shayancanonical: fixed in 105603853b0d9b3ba3f421ad9b34d4851eeef15f

src/connection_manager.py
shayancanonical: fixed in 105603853b0d9b3ba3f421ad9b34d4851eeef15f

src/connection_manager.py
shayancanonical: ah, yeah, we now raise an exception instead of None - thus we can guarantee return of list

105603853b0d9b3ba3f421ad9b34d4851eeef15f

src/connection_manager.py
shayancanonical: yep, now raise exception

105603853b0d9b3ba3f421ad9b34d4851eeef15f

src/connection_manager.py
shayancanonical: method was removed in 105603853b0d9b3ba3f421ad9b34d4851eeef15f

src/connection_manager.py
shayancanonical: nope, removed in 105603853b0d9b3ba3f421ad9b34d4851eeef15f

src/connection_manager.py
shayancanonical: updated in 105603853b0d9b3ba3f421ad9b34d4851eeef15f

src/connection_manager.py
shayancanonical: yes, there's no way i found to "tag" a connection with possible metadata. (may be we can try adding key/values to `extra-dejson`, but the naming of the connections help identify those created for dag bundles through relations with s3/git integrator)

tests/unit/test_connection_manager.py
```diff
@@ -0,0 +1,344 @@
...
+
+    def test_has_connection_for_git_changed_password(
+        self,
+        airflow_connection_manager,
+        valid_list_airflow_connections_output,
+        valid_git_provider_model,
+    ):
+        assert airflow_connection_manager.has_connection_for_git_changed(
+            "git_relation_1_connection", valid_s3_connection_info
+        )
```
jananisenthilkumar291, line 274: `valid_s3_connection_info` is referenced, but is not a fixture parameter?

tests/unit/test_connection_manager.py
```diff
@@ -0,0 +1,344 @@
...
+
+        valid_git_provider_model.password = "new-test-password"
```
jananisenthilkumar291: `valid_git_provider_model` does not have a `password` field.
It should perhaps be `credentials_personal_access_token`?

src/connection_manager.py
```diff
@@ -0,0 +1,295 @@
...
+        elif git_provider_model.authentication_method == git.AuthenticationMethodEnum.SSH:
+            # Existing airflow connection has credentials or
+            # SSH parameters changed
+            has_authentication_changed = (
+                airflow_connection.login
+                or airflow_connection.password
+                or airflow_connection.extra_dejson.get("private_key")
+                != git_provider_model.ssh_private_key
+                or json.loads(airflow_connection.extra_dejson.get("strict_host_key_checking"))
+                != git_provider_model.ssh_strict_host_key_checking
```
jananisenthilkumar291, line 222:
```suggestion
                or airflow_connection.extra_dejson.get("strict_host_key_checking")
                != git_provider_model.ssh_strict_host_key_checking
```

src/connection_manager.py
```diff
@@ -0,0 +1,295 @@
...
+                self._charm._command_executor.add_airflow_git_connection(
+                    connection_id,
+                    git_provider_model,
+                )
+
+                self.refresh()
+
+    def delete_stale_connections(self) -> None:
+        """Delete stale S3/git Airflow connections (those without relations)."""
+        if self.airflow_connections is None:
```
jananisenthilkumar291: Looks like `airflow-connections` never returns None. It's a dead code and can be removed.

src/connection_manager.py
```diff
@@ -0,0 +1,295 @@
...
+        )
+
+    def create_or_update_git_connections(self) -> None:
+        """Create or update git connections that have changed."""
+        airflow_connection_ids = [connection.conn_id for connection in self.airflow_connections]
+
+        for (
+            relation_id,
+            git_provider_model,
+        ) in self._charm._git_requires.get_git_connection_information().items():
```
jananisenthilkumar291, line 239: This defeats the purpose of charm having the `_git_requires` as a private property. 
A say `git_relation_connections` public property should be added to the charm for consistency (following `s3_relation_connections`).

src/command_executor.py
```diff
@@ -190,3 +191,56 @@ def list_airflow_connections(self) -> ops.pebble.ExecProcess:
...
             },
         )
+
+    @execute_pebble_exec_process
+    def add_airflow_git_connection(
+        self,
+        connection_id: str,
+        git_provider_model: git.GitProviderModel,
+    ) -> ops.pebble.ExecProcess:
+        """Add/update Airflow S3 connection.
```
jananisenthilkumar291:
```suggestion
        """Add/update Airflow git connection.
```

src/charm.py
```diff
@@ -363,13 +336,16 @@ def _validate_configs(self):
                     ops.BlockedStatus,
                 )
 
-    def _perform_potential_s3_connection_checks(self) -> None:
-        """Checks validity of all present s3 connections."""
+    def _perform_dag_bundle_connection_checks(self) -> None:
+        """Check validity of all present S3/git relations."""
+        if not self._s3_requires.relations and not self._git_requires.relations:
```
jananisenthilkumar291: The check does not seem to add substantial value as the checks are done later.

src/charm.py
```diff
@@ -381,6 +357,22 @@ def _perform_potential_s3_connection_checks(self) -> None:
                     ops.BlockedStatus,
                 )
 
+        if self._git_requires.relations:
+            errorneous_relation_ids = [
```
jananisenthilkumar291: Nit:
`errorneous_relation_ids` should be `erroneous_relation_ids`? Change here demands change in various places.

jananisenthilkumar291 (commented):
Left a few comments. Thanks @shayancanonical

src/connection_manager.py
shayancanonical: we need the json.loads() here to ensure proper comparison (because airflow connections return `extra-dejson` as a key/value pair of strings to strings

tests/unit/test_connection_manager.py
shayancanonical: oof, seems like this comment helped uncover a number of logical errors/bugs in the git related tests. thanks!

fixed in 08419c51e55d28ded01a8deb96096272bd38803c

tests/unit/test_connection_manager.py
shayancanonical: good catch. this is one of the many issues with the test fixed in 08419c51e55d28ded01a8deb96096272bd38803c

src/connection_manager.py
shayancanonical: true, a prior commit raised an error instead of returning None. seems like this was a leftover that wasnt cleaned up

removed in 08419c51e55d28ded01a8deb96096272bd38803c

src/connection_manager.py
shayancanonical: i like the idea of a consistent property. extracted in 08419c51e55d28ded01a8deb96096272bd38803c

src/command_executor.py
shayancanonical: fixed in 08419c51e55d28ded01a8deb96096272bd38803c

src/charm.py
shayancanonical: valid point. removed in 08419c51e55d28ded01a8deb96096272bd38803c

src/charm.py
shayancanonical: idk why the spellcheck didnt catch this. i'm getting really bad with misspells recently

fixed in 08419c51e55d28ded01a8deb96096272bd38803c

src/config_generator.py
DnPlas: I can't remember exactly this feedback, it was a while ago. Resolving for now.

DnPlas (approved):
Thanks @shayancanonical !

PR #51, opened 2026-04-15
[WF-346] Implement support for Airflow core configurations + add relevant scenario tests

shayancanonical:
## Description

- Add configuration options + default value changes for airflow configurations outlined in [WF022 - Airflow Core Configurability[(https://docs.google.com/document/d/1gON3iZXLvK5A3SdeyjW3bxUuQRs2v9gm6AhZoZaGPzg/edit?usp=sharing)
- Add scenario tests

---

---
## Checklist (all that apply)
- [ ] Documentation updated


src/config_generator.py
```diff
@@ -110,6 +112,50 @@ def api_server_config(self) -> dict[str, dict[str, str]]:
...
+        """Return the Airflow core config extracted from this charm's juju config.
+
+        This property also updates changes values of certain Airflow configurations
+        (those options that are not dynamically set to charm config values).
+
+        Uses the same {section: {key: value}} pattern as executor config.
+        """
+        return {
+            "api": {
+                "enable_swagger_ui": False,
```
DnPlas: I think these values can be directly added to the jinja2 template if we are not allowing changes.

DnPlas (commented):
Tiny comment, other than that LGTM

src/config_generator.py
shayancanonical: great call - moved the configs with fixed values to the jinja template in c08f3d0bb165dc0d10618fcee7a7d491b4512a4b

DnPlas (dismissed):
Thanks @shayancanonical !

src/charm.py
```diff
@@ -319,6 +321,38 @@ def _kubernetes_executor_pod_spec(self) -> str | None:
             return None
         return content.kubernetes_executor_pod_spec
 
+    def _validate_configs(self):
+        """Ensure validity of charm configurations."""
+        if not self.config:
```
jananisenthilkumar291: I believe it's a dead code as this seems to always evaluate to `True` as self.config return [ConfigData](https://documentation.ubuntu.com/ops/latest/reference/ops/#ops.CharmBase.config:~:text=it%20is%20running.-,property%20config%3A%20ConfigData,-%C2%B6) that is populated with defaults from `charmcraft.yaml`

src/config_generator.py
```diff
@@ -110,6 +112,42 @@ def api_server_config(self) -> dict[str, dict[str, str]]:
...
+            "core": {
+                "default_timezone": self._charm.config[constants.CORE_DEFAULT_TIMEZONE_CONFIG],
+                "max_active_runs_per_dag": self._charm.config[
+                    constants.CORE_MAX_ACTIVE_RUNS_PER_DAG_CONFIG
+                ],
+                "max_active_tasks_per_dag": self._charm.config[
+                    constants.CORE_MAX_ACTIVE_TASKS_PER_DAG_CONFIG
+                ],
+                "parallelism": self._charm.config[constants.CORE_PARALLELISM_CONFIG],
+                "default_impersonation": constants.WORKLOAD_USER,
```
jananisenthilkumar291: I believe the `default_impersonation` can also be moved to jinja template as it does not necessarily get populated by user input.

tests/unit/test_charm.py
```diff
@@ -960,6 +961,96 @@ def test_runtime_secret_created_when_peer_has_no_plaintext_fields(
         assert sensitive_data["api_auth__jwt_secret"]
 
 
+class TestAirflowConfigurability:
+    def test_negative_value_configs(self, context, state, mock_command_executor):
```
jananisenthilkumar291: could we consider `@pytest.mark.parameterize` over all 6 configs and not just `CORE_MAX_ACTIVE_RUNS_PER_DAG_CONFIG`

src/config_generator.py
```diff
@@ -110,6 +112,42 @@ def api_server_config(self) -> dict[str, dict[str, str]]:
             },
         }
 
+    @property
+    def coordinator_charm_core_config(self) -> dict[str, dict[str, str | bool]]:
+        """Return the Airflow core config extracted from this charm's juju config.
+
+        This property also updates changes values of certain Airflow configurations
+        (those options that are not dynamically set to charm config values).
```
jananisenthilkumar291: `updates changes` is garbled.
```suggestion
        This property also sets fixed operational values that are not user-configurable.
```

src/config_generator.py
```diff
@@ -110,6 +112,42 @@ def api_server_config(self) -> dict[str, dict[str, str]]:
             },
         }
 
+    @property
+    def coordinator_charm_core_config(self) -> dict[str, dict[str, str | bool]]:
```
jananisenthilkumar291: I see config types as either `string` or `int`
```suggestion
    def coordinator_charm_core_config(self) -> dict[str, dict[str, str | int]]:
```

src/charm.py
```diff
@@ -319,6 +321,38 @@ def _kubernetes_executor_pod_spec(self) -> str | None:
...
+                    config_name=constants.CORE_DEFAULT_TIMEZONE_CONFIG
+                ),
+                ops.BlockedStatus,
+            )
+
+        for config in [
+            constants.CORE_MAX_ACTIVE_RUNS_PER_DAG_CONFIG,
+            constants.CORE_MAX_ACTIVE_TASKS_PER_DAG_CONFIG,
+            constants.CORE_PARALLELISM_CONFIG,
+            constants.DAG_PROCESSOR_PARSING_PROCESSES_CONFIG,
```
jananisenthilkumar291: This `DAG_PROCESSOR_PARSING_PROCESSES_CONFIG` should perhaps be guarded against being set to '0'?
With this current check we're having an infinite _zombie_ process running during the infinite `while` loop [here](https://github.com/apache/airflow/blob/3.1.8/airflow-core/src/airflow/dag_processing/manager.py#L380).

src/charm.py
```diff
@@ -319,6 +321,38 @@ def _kubernetes_executor_pod_spec(self) -> str | None:
...
+            constants.DATABASE_SQL_ALCHEMY_POOL_SIZE_CONFIG,
+            constants.TRIGGERER_CAPACITY_CONFIG,
```
jananisenthilkumar291: Same goes with `TRIGGERER_CAPACITY_CONFIG`. Having this set to 0 will have a paralyzed triggerer and potentially (or actually) raise a ValueError [Refer](https://github.com/apache/airflow/blob/3.1.8/airflow-core/src/airflow/jobs/triggerer_job_runner.py#L123).

src/charm.py
```diff
@@ -583,9 +619,12 @@ def _reconcile(self, event: ops.EventBase) -> None:
...
                 self._config_generator.config_template_with_extra_config(
-                    **self._config_generator.api_server_config,
-                    **self._config_generator.dag_processor_config,
-                    **(self._kubernetes_executor_config or {}),
+                    **mergedeep.merge(
+                        self._config_generator.api_server_config,
+                        self._config_generator.dag_processor_config,
+                        (self._kubernetes_executor_config or {}),
+                        self._config_generator.coordinator_charm_core_config,
+                    ),
```
jananisenthilkumar291: I believe this updates `self._config_generator.api_server_config` inline? Instead can we use an empty dict as an accumulator?

```suggestion
                    **mergedeep.merge(
                        {},
                        self._config_generator.api_server_config,
                        self._config_generator.dag_processor_config,
                        (self._kubernetes_executor_config or {}),
                        self._config_generator.coordinator_charm_core_config,
                    ),
```

src/config_generator.py
```diff
@@ -110,6 +112,42 @@ def api_server_config(self) -> dict[str, dict[str, str]]:
...
+                "max_active_runs_per_dag": self._charm.config[
+                    constants.CORE_MAX_ACTIVE_RUNS_PER_DAG_CONFIG
+                ],
+                "max_active_tasks_per_dag": self._charm.config[
+                    constants.CORE_MAX_ACTIVE_TASKS_PER_DAG_CONFIG
+                ],
+                "parallelism": self._charm.config[constants.CORE_PARALLELISM_CONFIG],
+                "default_impersonation": constants.WORKLOAD_USER,
+            },
+            "dag_processor": {
```
jananisenthilkumar291, line 132: Any reason why this config cannot be moved into [day_processor_config](https://github.com/canonical/airflow-coordinator-k8s-operator/blob/feature/core_configurability/src/config_generator.py#L152)?
This could totally help in avoiding the dependency on `mergedeep`.

jananisenthilkumar291 (commented):
Thanks @shayancanonical, left some comments.

src/charm.py
shayancanonical: good call. removed in 664290ef4c37adc950d45249eda49bb7628fe3e4

src/config_generator.py
shayancanonical: we dont have a mechanism (or somehow refactored out the mechanism) to add parametrized configs to the jinja2 template that are not secret values

add `default_impersonation = ubuntu` to jinja template: a383374e6b9ea222b5a26eca337130351cc31ffa

tests/unit/test_charm.py
shayancanonical: added parametrized values in 664290ef4c37adc950d45249eda49bb7628fe3e4 to test all configs

src/config_generator.py
shayancanonical: removed sentence as this method no longer changes default values of certain configs 664290ef4c37adc950d45249eda49bb7628fe3e4

src/config_generator.py
shayancanonical: addressed in 664290ef4c37adc950d45249eda49bb7628fe3e4

src/charm.py
shayancanonical: disallow 0 as a valid value: 664290ef4c37adc950d45249eda49bb7628fe3e4

src/charm.py
shayancanonical: disallow 0 as a valid value: 664290ef4c37adc950d45249eda49bb7628fe3e4

src/charm.py
shayancanonical: avoid inline overrides in 664290ef4c37adc950d45249eda49bb7628fe3e4

src/config_generator.py
shayancanonical: i prefer not to have dicts per section of config file

rather hoped to create dicts that are logically grouped (e.g. configs from juju configs, configs for dag bundles, etc)

renamed the properties (e.g. `dag_processor_config` -> `dag_bundle_config`): 664290ef4c37adc950d45249eda49bb7628fe3e4

jananisenthilkumar291 (approved):
Thanks @shayancanonical

PR #58, opened 2026-05-01
[WF-381] updated db name TO BE parameterized

jananisenthilkumar291:
## Description

This PR aims to Update Database name to be parameterized.

---

---
## Checklist (all that apply)
- [ ] Documentation updated

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -56,7 +56,7 @@ class AirflowCoordinatorK8SOperatorCharm(ops.CharmBase):
         self._database_requires = data_interfaces_v0.DatabaseRequires(
             self,
             constants.POSTGRES_RELATION_NAME,
-            database_name=constants.AIRFLOW_DATABASE_NAME,
+            database_name=self.app.name.replace("-", "_"),
         )
         self._config_provider = airflow_coordinator.AirflowCoordinatorProvides(
             self,
--- a/src/charm.py
+++ b/src/charm.py
@@ -56,7 +56,7 @@ class AirflowCoordinatorK8SOperatorCharm(ops.CharmBase):
         self._database_requires = data_interfaces_v0.DatabaseRequires(
             self,
             constants.POSTGRES_RELATION_NAME,
-            database_name=self.app.name.replace("-", "_"),
+            database_name=f"{self.model.name}_{self.app.name}".replace("-", "_"),
         )
         self._config_provider = airflow_coordinator.AirflowCoordinatorProvides(
             self,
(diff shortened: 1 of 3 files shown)


src/charm.py
```diff
@@ -56,7 +56,7 @@ def __init__(self, framework: ops.Framework):
         self._database_requires = data_interfaces_v0.DatabaseRequires(
             self,
             constants.POSTGRES_RELATION_NAME,
-            database_name=constants.AIRFLOW_DATABASE_NAME,
+            database_name=f"{self.model.name}_{self.app.name}".replace("-", "_"),
```
shayancanonical: should we use the model uuid instead of the model name? it is possible to have 2 models of the same name across 2 different controllers + have cross-controller relations

using the model uuid will make the db name more unreadable compared to the model name. if we use model uuid, may be it should be a suffix instead of a prefix

jananisenthilkumar291: Done. Addressed in [1f97c52](https://github.com/canonical/airflow-coordinator-k8s-operator/pull/58/commits/1f97c526e287424af8e8d1be5ff40588a0a29306)

nikita-jare (commented):
`AIRFLOW_DATABASE_NAME` is now unused in `src/constants.py`
Can you also remove it?

src/constants.py
```diff
@@ -12,7 +12,7 @@
 PEER_RELATION_NAME = "coordinator-peers"
 
 POSTGRES_RELATION_NAME = "postgres"
-AIRFLOW_DATABASE_NAME = "airflow"
+# AIRFLOW_DATABASE_NAME = "airflow"
```
nikita-jare: Is there a reason this is commented out rather than deleted? If not, I think it should be removed.

nikita-jare (commented):
LGTM! Ready to approve.
Flagged a tiny nit below

src/constants.py
jananisenthilkumar291: Good Catch! That was a miss. Addressed in [54ec264](https://github.com/canonical/airflow-coordinator-k8s-operator/pull/58/commits/54ec26466734a4c5cf83ce5454502edbf7a9c0c7)

nikita-jare (dismissed):

shayancanonical (approved):

shayancanonical (changes requested):
on second thought, pg database names have a character limit of 63 characters

please ensure that the generated name does not exceed this limit

shayancanonical (approved):

PR #61, opened 2026-05-22
[WF-414] Update airflow coordinator charm lib to support optional distribution of webserver config

shayancanonical:
## Description

Implement changes to the airflow coordinator charm lib to optionally distribute the webserver config file (according to spec [WF-024 - Airflow Identity Integration](https://docs.google.com/document/d/16mzvH8-WNAC1Xu_WQrR89JlOQZmmS4psmW9l1vkiCAs/edit?usp=sharing) )

Changes include:
- AirflowCoordinatorProviderModel: adds optional field `webserver_config_template`
- AirflowCoordinatorProvides.set_airflow_config() accepts optional kwarg `webserver_config_template`
- AirflowCoordinatorCoreRequires: implements 3 methods to see if (1) webserver config can be written, (2) webserver config in workload containers needs update and (3) write webserver config file to workload container
- Associated unit tests for the modified charm lib methods

---

---
## Checklist (all that apply)
- [ ] Documentation updated

### DIFF
--- a/lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
+++ b/lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
@@ -275,16 +275,16 @@ class AirflowCoordinatorProviderModel(data_interfaces.BaseCommonModel):
     # (str from coordinator) and structured config dicts (from executor, deserialized
     # by data_interfaces' get_data). This will change when we refactor the library
     # to a much more generic one.
-    config_template: str | dict | None = None
-    kubernetes_executor_pod_spec: str | None = None
-    webserver_config_template: str | None = None
+    config_template: typing.Optional[str | dict] = None
+    kubernetes_executor_pod_spec: typing.Optional[str] = None
+    webserver_config_template: typing.Optional[str] = None
     sensitive_data: SensitiveDataSecretStr = None
-    secret_sensitive_data: data_interfaces.SecretString | None = None
+    secret_sensitive_data: typing.Optional[data_interfaces.SecretString] = None
 
-    validation_failures: str | None = None
+    validation_failures: typing.Optional[str] = None
 
     # CA chains for Airflow connections
-    tls_ca_chains: dict[str, str] | None = None
+    tls_ca_chains: typing.Optional[dict[str, str]] = None
 
     # hack to enable databag diff computation with data_interfaces v1 charm lib
     request_id: str = pydantic.Field(default="fixed_request_id", exclude=True)
@@ -322,8 +322,8 @@ class AirflowCoordinatorEvent(ops.EventBase, typing.Generic[TAirflowCoordinatorM
         self,
         handle: ops.Handle,
         relation: ops.Relation,
-        app: ops.Application | None,
-        unit: ops.Unit | None,
+        app: typing.Optional[ops.Application],
+        unit: typing.Optional[ops.Unit],
         content: TAirflowCoordinatorModels,
     ):
         super().__init__(handle)
@@ -666,7 +666,7 @@ class AirflowCoordinatorProviderEventHandler(
         self,
         config_template: str = None,
         kubernetes_executor_pod_spec: str = None,
-        webserver_config_template: str = None,
+        webserver_config_template: typing.Optional[str] = None,
         sensitive_data: dict[str, str] = {},
         tls_ca_chains: dict[str, str] = {},
     ):
--- a/lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
+++ b/lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
@@ -90,6 +90,12 @@ the pebble workload container
 to render and write the k8s executor pod spec
 8. `write_kubernetes_executor_pod_spec(filepath)`: renders and writes the k8s
 executor pod spec into the pebble workload container
+9. `can_write_webserver_config`: all prerequisities met to be able to render and
+write the webserver config file
+10. `webserver_config_needs_update(filepath)`: if the webserver config file in
+the relation has drifted from the contents of file on the workload container
+11. `write_webserver_config(filepath, user, group)`: write the webserver config
+file contents to the workload container
 
 `AirflowCoordinatorCoreRequires` will invoke the provided `callback` when:
 - the coordinator charm shares validation failures for all related core charms
@@ -165,7 +171,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 7
+LIBPATCH = 8
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
```diff
@@ -1006,6 +1017,46 @@ def write_kubernetes_executor_pod_spec(self, filepath: str, user: str, group: st
...
+        content = self._requirer_handler.provider_content
+        return bool(
+            self._ready
+            and content
+            and content.webserver_config_template
+            and content.sensitive_data
+        )
+
+    def webserver_config_needs_update(self, filepath: str) -> bool:
+        """Check whether the rendered webserver config differs from the file on disk."""
```
DnPlas: Tiny nit

```suggestion
        """Return True if the rendered webserver config differs from the file on disk; False otherwise."""
```

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
```diff
@@ -1006,6 +1017,46 @@ def write_kubernetes_executor_pod_spec(self, filepath: str, user: str, group: st
...
+
+        return on_disk != rendered
+
+    def write_webserver_config(self, filepath: str, user: str, group: str) -> None:
+        """Render and write webserver_config.py in the workload container."""
+        provider_content = self._requirer_handler.provider_content
+        rendered = jinja2.Template(provider_content.webserver_config_template).render(
+            **json.loads(provider_content.sensitive_data)
+        )
+        self._workload_container.push(filepath, rendered, user=user, group=group, make_dirs=True)
```
DnPlas: (non-blobking) Seems to me like we're now we can extract this into a helper method, this operation is repeated in `write_kubernetes_executor_pod_spec` and more or less on `write_tls_ca_chains` and `write_airflow_config`.

You may want to just create an abstraction for this.

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
```diff
@@ -1006,6 +1017,46 @@ def write_kubernetes_executor_pod_spec(self, filepath: str, user: str, group: st
...
+        return bool(
+            self._ready
+            and content
+            and content.webserver_config_template
+            and content.sensitive_data
+        )
+
+    def webserver_config_needs_update(self, filepath: str) -> bool:
+        """Check whether the rendered webserver config differs from the file on disk."""
+        provider_content = self._requirer_handler.provider_content
```
DnPlas, line 1041: I think you could add some guardrails here to check for when the `povider_content`'s attributes are `None`. I recommend we do something similar to `can_write_webserver_config`.

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
DnPlas: One thing you could also do is create a property for the rendered template (any) and use it here (:

tests/unit/test_airflow_coordinator.py
```diff
@@ -1033,6 +1036,159 @@ def test_write_tls_ca_chains_to_workload_container(
...
+                if expected_error:
+                    with pytest.raises(expected_error):
+                        manager.charm.requirer.webserver_config_needs_update("/test/file")
+                else:
+                    assert (
+                        manager.charm.requirer.webserver_config_needs_update("/test/file")
+                        == update_required
+                    )
+
+    def test_wrtie_webserver_config(
```
DnPlas: (: 

```suggestion
    def test_write_webserver_config(
```

DnPlas (commented):
Tiny comments, thanks @shayancanonical !

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
shayancanonical: We had similar guardrails in the past, but I recall a tech talk where we decided that we would implement methods/properties in the charm lib that have strong opinions on how they will be invoked (e.g. no consuming charm should cann `write_webserver_config` before checking `can_write_webserver_config` beforehand, where the `None` checks are performed)

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
shayancanonical: addressed in 59e40eef046382e020cce58d63d94acbdbc0d58a

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
shayancanonical: extracted property to render webserver config (since this was being done twice in the added code): 59e40eef046382e020cce58d63d94acbdbc0d58a

avoiding abstraction for workflow to render some jinja template and push to workload container because i think the overhead is more expensive than the gains (as it stands, the code is just 3 lines, even if repeated in 3 methods. with abstraction, one more hop would be added to someone tracing the the code and maintainability gets harder if one of the methods now needs to do something additional in the future)

tests/unit/test_airflow_coordinator.py
shayancanonical: addressed in 59e40eef046382e020cce58d63d94acbdbc0d58a

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
```diff
@@ -1185,6 +1241,7 @@ def set_airflow_config(
         self,
         config_template: typing.Optional[str] = None,
         k8s_executor_pod_spec_template: typing.Optional[str] = None,
+        webserver_config_template: typing.Optional[str] = None,
```
jananisenthilkumar291, line 1244: L280, L669 and here refer to the same concept. I believe we should pick one style and stick to it.

tests/unit/test_airflow_coordinator.py
```diff
@@ -1033,6 +1036,159 @@ def test_write_tls_ca_chains_to_workload_container(
...
+                True,
+                io.StringIO("content1"),
+                jinja2.exceptions.TemplateSyntaxError("test-error", lineno=1),
+                None,
+                jinja2.exceptions.TemplateSyntaxError,
+                id="error_rendering_relation_content",
+            ),
+        ],
+    )
+    def test_webserver_config_needs_update_unchanged_contents(
```
jananisenthilkumar291: The test covers 6 parameterized branches uincluding different-content and error cases. The name implies it only covers the no-change path. Consider updating the test name?

jananisenthilkumar291 (commented):
Added tiny nits, Rest LGTM!

tests/unit/test_airflow_coordinator.py
shayancanonical: addressed in 3387c0bad620ea3e1b08c006ed4c77654c1c1ea7

lib/charms/airflow_coordinator_k8s/v0/airflow_coordinator.py
shayancanonical: converged to using `typing.Optional` in 3387c0bad620ea3e1b08c006ed4c77654c1c1ea7

jananisenthilkumar291 (approved):
LGTM! Thanks @shayancanonical

PR #62, opened 2026-05-27
[WF-427] Implement relation endpoint for integration with hydra

shayancanonical:
## Description

- Adds new `oauth` relation endpoint
- Import hydra oauth charm lib
- Uses oauth charm lib requirer class to initiate integration with hydra if integration added
- Implements property to indicate active relation with hydra
- Adds unit test stub (to be further refined as the feature is incrementally implemented) 

---

---
## Checklist (all that apply)
- [ ] Documentation updated


src/charm.py
```diff
@@ -13,6 +13,7 @@
 import charms.airflow_coordinator_k8s.v0.airflow_coordinator as airflow_coordinator
 import charms.data_platform_libs.v0.data_interfaces as data_interfaces_v0
 import charms.git_integrator.v0.git as git
+import charms.hydra.v0.oauth as oauth
```
nikita-jare, line 16: this hydra lib is not vendored in the branch so unit and integration test are both failing at import.

src/charm.py
```diff
@@ -324,6 +331,20 @@ def _kubernetes_executor_pod_spec(self) -> str | None:
...
 
+    @property
+    def _oauth_active(self) -> bool:
+        """Return True when the oauth relation has valid provider credentials."""
+        if not self.model.get_relation(constants.OAUTH_ENDPOINT_NAME):
+            return False
+
+        try:
+            provider_info = self._oauth_requirer.get_provider_info()
+        except Exception:
```
nikita-jare, line 342: Nit: `get_provider_info()` returns None when the relation isn't ready, so this except Exception seems extra or could narrow to the specific exception the lib raises

pyproject.toml
```diff
@@ -32,6 +32,7 @@ unit = [
     "ops[testing]",
     "pytest",
     "cryptography>=41.0",
+    "jsonschema>=4.26.0",
```
nikita-jare: jsonschema is a [runtime dep](https://github.com/canonical/hydra-operator/blob/main/lib/charms/hydra/v0/oauth.py#L72) of the hydra lib, so it needs to be in [project].dependencies otherwise it won't be packed into the charm.

nikita-jare (commented):
The description says "Adds unit test stub" but there's no test file in the diff. Did the stub get left uncommitted along with the hydra lib?

src/charm.py
shayancanonical: honestly, `get_provider_info()` can raise a variety of errors (RuntimeError, any ops secret related errors, custom DataValidationError) - I agree that permission exception handling is bad practice, but I feel that I'd rather do it for this one statement than trace and enumerate all the possible exceptions (as the lib may introduce another one tomorrow that we would not be handling)

src/charm.py
shayancanonical: added in 46ebb910698bc1d66ae9a67f1fbb73bd8fdd22c0 - i somehow missed tracking newly added files

pyproject.toml
shayancanonical: great catch! moved to charm deps instead in 46ebb910698bc1d66ae9a67f1fbb73bd8fdd22c0

shayancanonical:
had missed tracking the newly added unittest file with stub tests. added in 46ebb910698bc1d66ae9a67f1fbb73bd8fdd22c0

src/charm.py
jananisenthilkumar291: Some exceptions like `DataValidationError` means a misconfigured provider(eg. malformed OIDC endpoints), which will look like "no relation" rather than surfacing a meaningful message to the operator. 
To reach a middle ground, could we catch exception, log it and set a Blobkedstatus so the operator knows that something is wrong?

tests/unit/test_oauth.py
```diff
@@ -0,0 +1,60 @@
...
+
+"""Tests for the OAuth integration."""
+
+import dataclasses
+
+import ops
+import ops.testing
+import pytest
+
+OAUTH_CLIENT_ID = "grafana_client_id"
```
jananisenthilkumar291:
```suggestion
OAUTH_CLIENT_ID = "airflow_client_id"
```

src/charm.py
nikita-jare: Fair point @shayancanonical, a broader catch makes sense.

@jananisenthilkumar291 agreed there's a gap. We can flag it from _reconcile based on the relation being present but not active, without classifying each error so the broad catch can stays as-is. Let's settle the exact status when _oauth_active gets used and track in that PR.

shayancanonical: will revisit once `_oauth_active` gets used in a followup PR (which i'll open later today)

tests/unit/test_oauth.py
shayancanonical: addressed in 54019c82ade9f75c8072a737e27eac271b7a5ab6

nikita-jare (approved):
Looks Good. Thank you Shayan!

PR #64, opened 2026-05-28
[WF-428] Add config options to support variability of identity integration

shayancanonical:
## Description

- Add configuration options related to the identity integration
- Add stub of unittests to be further refined as the feature is implemented

---

---
## Checklist (all that apply)
- [ ] Documentation updated


charmcraft.yaml
```diff
@@ -139,3 +139,32 @@ config:
...
+      description: |
+        Comma delimited list of groups in the external identity provider that maps to the Airflow viewer role
+    idp_groups_for_public:
+      type: string
+      default: ""
+      description: |
+        Comma delimited list of groups in the external identity provider that maps to the Airflow public role
+    enable_user_registration:
+      type: boolean
+      default: true
```
nikita-jare: We are defaulting this to True because this is an OIDC flow right?
As FAB defaults this to `False` [AUTH_USER_REGISTRATION = False](https://github.com/dpgaspar/Flask-AppBuilder/blob/master/flask_appbuilder/security/manager.py#L232).

nikita-jare (dismissed):
LGTM.
Can you update the title with relevant JIRA ticket to track?

charmcraft.yaml
shayancanonical: good point, changed the default to `false` as it is more restrictive - we can ask daniela or rob at the next standup if the default should be `true` instead. 361b0193122265d1b811309171363323d92098a2

nikita-jare (approved):

PR #66, opened 2026-06-03
[WF-429] Generate and distribute webserver_config.py when oauth integration created

shayancanonical:
## Description

- Builds upon PR #61, #62 and #64. This PR wires all the pieces together to generate a `webserver_config.py` file from a jinja template and distribute the generated file over `airflow-coordinator` relation interface to all downstream airflow core charms (with continued share of sensitive data over juju secrets)
- Sets the auth_manager to fab auth manager in the `airflow.cfg` when valid oauth relation formed
- Introduces a new file in `src/` to encapsulate logic around `webserver_config.py` generation
- Add scenario tests for behavior related to oauth integration + idp configs 

---

---
## Checklist (all that apply)
- [ ] Documentation updated


src/webserver_config_generator.py
```diff
@@ -0,0 +1,161 @@
...
+            logger.warning(
+                "OauthProviderConfig missing required fields: %s — cannot render webserver config",
+                missing,
+            )
+            raise WebserverConfigError("Missing OauthProviderConfig fields")
+
+        api_base_url = self._charm._config_generator._api_server_base_url
+        if not api_base_url:
+            logger.warning("api_base_url is empty; cannot render webserver config template")
+            return WebserverConfigError("api_server_base_url is empty")
```
nikita-jare: This should be `raise` not `return` which will be always truthy

src/charm.py
```diff
@@ -551,6 +558,22 @@ def _run_db_migrate(self) -> None:
                 constants.DB_MIGRATION_FAILED_MESSAGE, ops.BlockedStatus
             )
 
+    def _update_oauth_config(self) -> None:
+        """Update OAuth client config if relation exists."""
+        if self.model.get_relation(constants.OAUTH_ENDPOINT_NAME):
+            try:
+                self._oauth_requirer.update_client_config(
+                    oauth.ClientConfig(
+                        redirect_uri=self._config_generator._api_server_base_url,
```
nikita-jare: FAB's OAuth callback is `/oauth-authorized/<provider>` ([FAB views.py:709](https://github.com/dpgaspar/Flask-AppBuilder/blob/master/flask_appbuilder/security/views.py#L709)), So Hydra needs `<base_url>/oauth-authorized/hydra` registered, otherwise it'll reject the actual callback. Should be `{self._config_generator._api_server_base_url}/oauth-authorized/hydra`

src/constants.py
```diff
@@ -54,6 +54,7 @@
 DAG_PROCESSOR_PARSING_PROCESSES_CONFIG = "dag_processor_parsing_processes"
 DATABASE_SQL_ALCHEMY_POOL_SIZE_CONFIG = "database_sql_alchemy_pool_size"
 TRIGGERER_CAPACITY_CONFIG = "triggerer_capacity"
+ENABLE_USER_REGISTRATION_CONFIG = "enable_user_registration"
```
nikita-jare, line 57: This is defined twice. Same constant already declared at https://github.com/canonical/airflow-coordinator-k8s-operator/pull/66/changes#diff-42d0f00c4780305311eae71fecca993a0523587cb4156e039cabd0d1ad8dbbb8L72

tox.ini
```diff
@@ -59,7 +59,7 @@ commands =
     uv run coverage run \
         --source={[vars]src_path},{[vars]lib_path} --module pytest \
         {[vars]tests_path}/unit \
-        {posargs}
+        {posargs} -vv
```
nikita-jare: Do we need the -vv flag? Output will be noisy.

src/templates/webserver_config.py.j2
```diff
@@ -0,0 +1,32 @@
...
+AUTH_ROLES_MAPPING = {{ auth_roles_mapping }}
+
+# Hydra OAuth Provider Configuration
+OAUTH_PROVIDERS = [
+    {
+        "name": "hydra",
+        "icon": "fa-key",
+        "token_key": "access_token",
+        "remote_app": {
+            "client_id": "{{ client_id }}",
```
nikita-jare: Suggestion: drop the outer quotes and change to something like {{ client_id | tojson }} 
If any chars like newline or spaces are in the string it will break the rendered file.

Same for lines 23, 25, 27, 28, 29, and `auth_roles_mapping` on line 12

src/templates/webserver_config.py.j2
```diff
@@ -0,0 +1,32 @@
+from flask_appbuilder.security.manager import AUTH_OAUTH
+
+# Set Authentication Type to OAuth
+AUTH_TYPE = AUTH_OAUTH
+AUTH_ROLES_SYNC_AT_LOGIN = True
+
+# Enable user registration in addition to authentication with OAuth
+AUTH_USER_REGISTRATION = {{ enable_user_registration }}
+AUTH_USER_REGISTRATION_ROLE = "Viewer"
```
nikita-jare: FAB's default here is "Public" ([manager.py](https://github.com/dpgaspar/Flask-AppBuilder/blob/master/flask_appbuilder/security/manager.py#L234)). "Viewer" means any user the IdP authenticates, even if their groups don't match anything in AUTH_ROLES_MAPPING gets read-only access to everything. Worth either defaulting to "Public" or exposing as a config option?

src/charm.py
```diff
@@ -340,8 +344,10 @@ def _oauth_active(self) -> bool:
         try:
             provider_info = self._oauth_requirer.get_provider_info()
         except Exception:
-            logger.exception("Error retrieving OAuth provider info")
-            return False
+            raise ExceptionWithStatusError(
```
nikita-jare: Wondering if it'd still be cleaner to keep this returning False and do the status check in _reconcile?

nikita-jare (commented):
Looks like its coming together now.
Left a few inline comments and suggestions below.

src/webserver_config_generator.py
shayancanonical: oof, thanks for catching. updated in 444666896880c80c926a86c13fc48f3e083b9b5a

src/charm.py
shayancanonical: i'm not entirely sure, but this looks ok to me. updated in 444666896880c80c926a86c13fc48f3e083b9b5a

will validate when testing this week and open followup PR if this needs to be corrected

src/constants.py
shayancanonical: result of merge/rebase, thanks for catching. removed in 444666896880c80c926a86c13fc48f3e083b9b5a

tox.ini
shayancanonical: removed in 444666896880c80c926a86c13fc48f3e083b9b5a. mistaken inclusion in PR

src/templates/webserver_config.py.j2
shayancanonical: updated to `public` in 444666896880c80c926a86c13fc48f3e083b9b5a. public seems sensible, and im all for staying aligned with upstream

src/charm.py
shayancanonical: good catch. we shouldnt raise exceptions in a property used through the code base

added an explicit check in `_perform_checks()` which ensures validity of relations to catch this exception first thing in the reconcile method (before any logic run that can affect/update state of unit): 444666896880c80c926a86c13fc48f3e083b9b5a

src/templates/webserver_config.py.j2
shayancanonical: good suggestion. updated in 444666896880c80c926a86c13fc48f3e083b9b5a

auth_roles_mapping updated in 81366d3562352458cf0be2a3354a463833c69ffd

src/charm.py
```diff
@@ -551,6 +565,22 @@ def _run_db_migrate(self) -> None:
                 constants.DB_MIGRATION_FAILED_MESSAGE, ops.BlockedStatus
             )
 
+    def _update_oauth_config(self) -> None:
+        """Update OAuth client config if relation exists."""
+        if self.model.get_relation(constants.OAUTH_ENDPOINT_NAME):
+            try:
+                self._oauth_requirer.update_client_config(
+                    oauth.ClientConfig(
+                        redirect_uri=f"{self._config_generator._api_server_base_url}/oauth-authroized/hydra",
```
nikita-jare:
```suggestion
                        redirect_uri=f"{self._config_generator._api_server_base_url}/oauth-authorized/hydra",
```

nikita-jare (commented):
Ready to approve except for the tiny typo below.

src/charm.py
shayancanonical: addressed in 02a1a7c44f53f4adcc8c5efee43642d03705d629

nikita-jare (approved):
Thank you @shayancanonical  :)

shayancanonical:
thanks for the reviews

